PR #7, opened 2015-11-09
channel.open on opened channel should throw channel_error

deadtrickster:
All 23 tests passed.

### DIFF
--- a/src/rabbit_channel.erl
+++ b/src/rabbit_channel.erl
@@ -855,7 +855,7 @@ handle_method(#'channel.open'{}, _, State = #ch{state = starting}) ->
 
 handle_method(#'channel.open'{}, _, _State) ->
     rabbit_misc:protocol_error(
-      command_invalid, "second 'channel.open' seen", []);
+      channel_error, "second 'channel.open' seen", []);
 
 handle_method(_Method, _, #ch{state = starting}) ->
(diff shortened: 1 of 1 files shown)


michaelklishin:
Originally submitted as https://github.com/rabbitmq/rabbitmq-website/pull/106. This change is breaking but double-opening a channel is an unrecoverable operation, and most apps should (and do) treat it as a generic fatal error. Therefore we should do what the spec says.

michaelklishin:
@deadtrickster FYI, most of the tests are actually in `rabbitmq_test`.

deadtrickster:
offtopic: how to override rabbitmq-common path? just break rabbitmq-components.mk magic and replace path?

michaelklishin:
@deadtrickster temporarily modifying `rabbitmq-components.mk` is perhaps the easiest route. Maybe @dumbbell can recommend an erlang.mk feature that I'm not familiar with, though.

dumbbell:
@deadtrickster: If you mean the path on disk, erlang.mk expects all dependencies in the same directory. You can change this directory by setting `DEPS_DIR` to whatever you want.

If you mean the URL of the repository, you should be able to set `dep_rabbit_common` to what you need (for example, `dep_rabbit_common = git https://github.com/deadtrickster/rabbitmq-common.git master`). You probably need to add this between `rabbitmq-components.mk` inclusion and `erlang.mk` one.

PR #8, opened 2015-11-10
Use Ranch for TCP acceptor pool

essen:
Part of https://github.com/rabbitmq/rabbitmq-server/issues/260.

I didn't add a dependency on Ranch yet; I suspect rabbit_common
is used by more than the server, and we would depend on Ranch
even for unrelated things. I will need to check this before
merging.

### DIFF
--- a/src/rabbit_reader.erl
+++ b/src/rabbit_reader.erl
@@ -53,12 +53,12 @@
 -include("rabbit_framing.hrl").
 -include("rabbit.hrl").
 
--export([start_link/1, info_keys/0, info/1, info/2, force_event_refresh/2,
+-export([start_link/3, info_keys/0, info/1, info/2, force_event_refresh/2,
          shutdown/2]).
 
 -export([system_continue/3, system_terminate/4, system_code_change/4]).
 
--export([init/2, mainloop/4, recvloop/4]).
+-export([init/4, mainloop/4, recvloop/4]).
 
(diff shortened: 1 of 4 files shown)


michaelklishin:
`rabbitmq_common` is shared by the server and our Erlang client.

michaelklishin:
Testing GH notifications.

michaelklishin:
Ditto.

michaelklishin:
@essen the change doesn't seem to add a dependency on Ranch. Is this intentional?

michaelklishin:
OK, `rabbit` does depend on Ranch.

essen:
Yeah I'm not sure we should depend on Ranch from common, considering it's also used by the client which does not (and should not have to) need Ranch.

src/rabbit_reader.erl
```diff
@@ -340,13 +345,11 @@ start_connection(Parent, HelperSup, Deb, Sock, SockTransform) ->
            end,
     {ok, HandshakeTimeout} = application:get_env(rabbit, handshake_timeout),
     InitialFrameMax = application:get_env(rabbit, initial_frame_max, ?FRAME_MIN_SIZE),
-    ClientSock = socket_op(Sock, SockTransform),
-    erlang:send_after(HandshakeTimeout, self(), handshake_timeout),
```
michaelklishin: @essen how are handshake timeouts handled now?

PR #13, opened 2015-11-11
More sensible error messages for connection-level exceptions on running connections

michaelklishin:
Fixes https://github.com/rabbitmq/rabbitmq-server/issues/419, includes #9, #10, #11, #12.

### DIFF
--- a/src/rabbit_reader.erl
+++ b/src/rabbit_reader.erl
@@ -787,11 +787,7 @@ handle_exception(State = #v1{connection = #connection{protocol = Protocol,
         "Error on AMQP connection ~p (~s,"
         " user: '~s', state: ~p):~n~s~n",
         [self(), ConnName, User#user.username, opening, ErrMsg]),
-    {0, CloseMethod} =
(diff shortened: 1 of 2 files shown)


src/rabbit_reader.erl
```diff
@@ -759,6 +743,30 @@ maybe_close(State) ->
 termination_kind(normal) -> controlled;
 termination_kind(_)      -> uncontrolled.
 
+format_hard_error(#amqp_error{name = N, explanation = E, method = M}) ->
+    io_lib:format("operation ~s caused a connection exception ~s: ~p", [M, N, E]);
+format_hard_error(Reason) ->
+    Reason.
+
+log_connection_exception(Name, Ex) ->
+  Severity = case Ex of
```
videlalvaro: since you are refactoring, why not reformat the case here? either move it to a new line or add the Emacs spacing?

michaelklishin: Sure, I've forgotten to run formatting before committing.

PR #16, opened 2015-11-17
Makes connection_closed_abruptly exception more informative.

Ayanda-D:
Fixes rabbitmq/rabbitmq-server#60.

### DIFF
--- a/src/rabbit_reader.erl
+++ b/src/rabbit_reader.erl
@@ -401,20 +401,18 @@ start_connection(Parent, HelperSup, Deb, Sock, SockTransform) ->
     done.
 
 log_connection_exception(Name, Ex) ->
-    Severity = case Ex of
-                   connection_closed_with_no_data_received -> debug;
-                   connection_closed_abruptly              -> warning;
-                   _                                       -> error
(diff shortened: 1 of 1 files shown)


michaelklishin:
OK, so this is a fairly cosmetic revision of https://github.com/rabbitmq/rabbitmq-common/commit/bffededc34d5db793988ee13f5bdb2ffeee3a9e4, am I correct. It's true that I forgot to close https://github.com/rabbitmq/rabbitmq-server/issues/60.

I'm not against this version but looks like we did the same change twice. Originally the error message had an Erlang tuple logged and it made little sense to most people.

Ayanda-D:
@michaelklishin okay cool. also didnt notice this had been dealt with in https://github.com/rabbitmq/rabbitmq-common/commit/bffededc34d5db793988ee13f5bdb2ffeee3a9e4. so n/a. :+1:

michaelklishin:
@Ayanda-D I think we should close this. Not to say that my solution is superior in any way, it's just a purely cosmetic change at this point. Thank you for your time!

PR #22, opened 2015-11-20
Restore the handshake timeout functionality in rabbit_reader

### LINKED ISSUE #21: Handshake timeout was incorrectly removed
The Ranch patch incorrectly removed this feature. It needs to be restored.

essen:
Fixes #21.


essen:
I ran "make unit" to confirm everything works.

michaelklishin:
:+1:

PR #26, opened 2015-11-20
Add retry limit to rabbit_amqqueue:with/2,3

binarin:
Looks like rabbit_mnesia:is_process_alive/1 check mentioned in comments
there is not sufficient - there was a rabbitmq instance that spent more
than 2 days in this loop.

It should improve situation described in https://github.com/rabbitmq/rabbitmq-server/issues/349 - in some reasonable amount of time client will be unblocked and crash report will be logged.

### DIFF
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -397,6 +397,11 @@ not_found_or_absent_dirty(Name) ->
     end.
 
 with(Name, F, E) ->
+    with(Name, F, E, 1000).
+
+with(_Name, _F, E, 0) ->
+    E(infinite_loop_detected);
+with(Name, F, E, RetriesLeft) ->
     case lookup(Name) of
         {ok, Q = #amqqueue{state = crashed}} ->
             E({absent, Q, crashed});
@@ -410,7 +415,7 @@ with(Name, F, E) ->
             rabbit_misc:with_exit_handler(
               fun () -> false = rabbit_mnesia:is_process_alive(QPid),
                         timer:sleep(25),
-                        with(Name, F, E)
+                        with(Name, F, E, RetriesLeft - 1)
               end, fun () -> F(Q) end);
         {error, not_found} ->
             E(not_found_or_absent_dirty(Name))


michaelklishin:
Agreed. 20-30 seconds is the timeout value we've settled on for socket operations in many clients, too.

src/rabbit_amqqueue.erl
```diff
@@ -397,6 +397,11 @@ not_found_or_absent_dirty(Name) ->
     end.
 
 with(Name, F, E) ->
+    with(Name, F, E, 1000).
+
+with(_Name, _F, E, 0) ->
+    E(infinite_loop_detected);
```
michaelklishin: Should we make `with_or_die/2` handle this case gracefully? at least with something that other `function_clause`?

binarin: I don't think it's useful - this is still about a bug in some other part of the system. This patch will just allow to detect this condition earlier. Maybe knowing exact time will help to localize this bug.

michaelklishin: It is useful for clients to get a sensible error messages instead of `internal error`.

binarin:
Sadly this is useless currently.
Now another infinite loop happens with the help of this line - https://github.com/rabbitmq/rabbitmq-common/blob/rabbitmq_v3_6_0/src/rabbit_channel.erl#L1346

I'm not quite sure, but looking at functions used by `with/4` I see some mismatch:
1) `rabbit_amqqueue:lookup/1` reads `rabbit_queue` table
2) `rabbit_amqqueue:not_found_or_absent_dirty/1` reads `rabbit_durable_queue`
Not sure is something needs to be fixed there.

And the primary question is https://github.com/rabbitmq/rabbitmq-common/blob/rabbitmq_v3_6_0/src/rabbit_amqqueue.erl#L417 - it sends a wrong signal to the other parts of the code - 'not_found' is something mild, when the actual message should be "something bad has happened to the queue, but we don't know what it is". So maybe returning `internal error` to the client is indeed the best approach here.

michaelklishin:
@binarin there are two tables for queues. When a non-mirrored queue hosted on an unavailable node is accessed, for durable queues we respond with `not_found`. This is by design. `internal_error` isn't precise and guarantees we will be getting endless support requests about it.

I guess we could extend the protocol with a new code that would mean "currently unavailable" — is this what you are after?

binarin:
What do you think about something like this - https://github.com/binarin/rabbitmq-common/commit/f39f3faeb60d7fe74c9a65e9a20249c43c974419 ?

michaelklishin:
@binarin the idea is reasonable, the message might need some clarification.

PR #31, opened 2015-12-01
Detect connections dropped during resource alarm

binarin:
Fixes https://github.com/rabbitmq/rabbitmq-server/issues/384 in the only
possible (portable) way - heartbeat sender detects dead socket during
normal operation and notifies rabbit_reader about this fact. This is
especially good given the recent decrease of default heartbeat timeout
to 60 seconds.

This way dead socket will be detected at most in 2 heartbeat
timeouts (as the first write to dropped socket will actually succeed).

Other probable fixes are non-portable or probably not worth it:
- Some unix versions allow to get a TCP connection state using
  getsockopt, but there is no such option for Windows.
- 'netstat' output could be parsed periodically, but I'm not sure that
  the overhead is comparable with a problem we are trying to solve.

### DIFF
--- a/src/rabbit_reader.erl
+++ b/src/rabbit_reader.erl
@@ -489,6 +489,10 @@ mainloop(Deb, Buf, BufLen, State = #v1{sock = Sock,
             stop(tcp_healthcheck, State);
         closed ->
             stop(closed, State);
+        {other, {heartbeat_send_error, Reason}} ->
+            %% The only portable way to detect disconnect on blocked
+            %% connection is to wait for heartbeat send failure.
+            stop(Reason, State);
         {error, Reason} ->
             stop(Reason, State);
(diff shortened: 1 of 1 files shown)


michaelklishin:
This looks reasonable but there is one important aspect we need to consider: with many clients, a blocked connection won't be considered to be dead due to heartbeats. It's debatable whether it should be
but that's certainly the case for some clients.

If this is merged, with lower-ish heartbeat intervals (e.g. 5-10 seconds, which is not uncommon), connections would be dropped by the server fairly quickly when it encounters a resource-based alarm.

Is this nonetheless a better outcome than what we have today for end user?

michaelklishin:
Some experiments suggest that I may be wrong: only connections of disconnected clients end up being cleaned up, which is what we want.

PR #32, opened 2015-12-04
Unsigned field types to support AMQP-0-9-1

hairyhum:
Added few unsigned types for better compatibility with AMQP-0-9-1
https://github.com/rabbitmq/rabbitmq-server/issues/20

### DIFF
--- a/codegen.py
+++ b/codegen.py
@@ -351,6 +351,7 @@ def genErl(spec):
 
 -type(amqp_field_type() ::
       'longstr' | 'signedint' | 'decimal' | 'timestamp' |
+      'unsignedbyte' | 'unsignedshort' | 'unsignedint' |
       'table' | 'byte' | 'double' | 'float' | 'long' |
       'short' | 'bool' | 'binary' | 'void' | 'array').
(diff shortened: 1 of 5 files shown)


deadtrickster:
Hey! any plans for unsigned long? (uint64). what letter it could be assigned to?

michaelklishin:
@deadtrickster is there really much use in supporting unsigned longs in attribute tables? FWIW most popular languages today don't support (or see widely used) unsigned anything these days.

deadtrickster:
@michaelklishin Right now I can came up only with 64-bit pointers and monitoring examples. Pointers probably not relevant here :-) But say I want to monitor server(s) (or other monitorable thing) including memory and network usage. Values probably exceed uint32 very quickly. Of course data can go through message body but what if I also have nice custom header-like exchange that routs bad behaved stuff to emergency queue :-)

hairyhum:
@deadtrickster there can be confusion if we add unsigned long, because it's field type tag already used for signed long (See http://www.rabbitmq.com/amqp-0-9-1-errata.html).

PR #34, opened 2015-12-04
fix failed integer value negotiation reporting

deadtrickster:
BEFORE: negotiated frame_max = 1024 is lower than the minimum allowed value (131072)
AFTER: negotiated frame_max = 1024 is lower than the minimum allowed value (4096)

Sorry last one was mistakenly closed


michaelklishin:
I will cherry-pick to stable => milestone `3.5.7`.

PR #35, opened 2015-12-14
Generate pseudo-random delivery tags

hairyhum:
Related to https://github.com/rabbitmq/rabbitmq-server/issues/255


michaelklishin:
This is a misunderstanding of one option in #255. We can't break how delivery tags are generated:

 * `basic.ack` with `multiple` = true depends on acks being incrementing integers
 * this is how things have been working since day 1, so for almost 9 years now.
 * this would make consumer delivery tags to be generated in a different way from publisher confirms, and the latter cannot be updated as it would break all existing clients.

Issue #36, opened 2015-12-17
Codegen scripts don't work with Python 3


essen:
Many systems today come with Python 3 being the default "python". But the codegen scripts don't work with it. We should support both Python 2 and 3.

Patch incoming.

essen:
Ready to be reviewed.

dumbbell:
I tested the two pull requests and it's working for me too. I compared the files generated with Python 3 to the files generated with Python 2 without these branches, and there is no difference (beside an unordered list with a different order).

There are two more places needing a similar patch:
* rabbitmq-amqp1.0
* rabbitmq-java-client

I will work on them.

essen:
:birthday:

PR #37, opened 2015-12-17
Make codegen.py work with Python 3

essen:
Part of https://github.com/rabbitmq/rabbitmq-common/issues/36


michaelklishin:
Did some sanity checking with Python 2, LGTM.

PR #41, opened 2016-01-04
Move ch record definition to rabbit.hrl

deadtrickster:
Follow up pull request for this discussion: https://groups.google.com/forum/#!topic/rabbitmq-users/fJM_Y71jAQ4

Nothing is expected to broke since rabbit_channel includes rabbit.hrl already


michaelklishin:
@dumbbell do you have any objections to this?

dumbbell:
Hi @deadtrickster!

What do you need this record for?

For instance, would it be ok if we provided functions to access members of this record, instead of making it public?

deadtrickster:
@dumbbell Hey!

For my plugin I need to access channel vhost: https://github.com/deadtrickster/rabbitmq_logs_exchange/blob/master/src/rabbit_message_id_interceptor.erl#L103

Just curious why don't make it public?

videlalvaro:
If you need to access the vhost you have this function: https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_channel.erl#L2026 which was exported specifically for interceptors

michaelklishin:
Yes, if all you need is a single field then there is no point in moving the entire record.

PR #42, opened 2016-01-08
Use dynamic code loading in ssl_compat and time_compat

dcorbacho:
Based on dynamic code load in startup/runtime. See https://github.com/rabbitmq/rabbitmq-server/issues/346#issuecomment-169980207

Solves https://github.com/rabbitmq/rabbitmq-server/issues/346 and https://github.com/rabbitmq/rabbitmq-server/issues/347

### DIFF
--- a/src/code_version.erl
+++ b/src/code_version.erl
@@ -33,7 +33,10 @@
 %% in ERTS 7.0, while providing compatibility with previous versions.
 %%
 %% `Module` must contain an attribute `erlang_version_support` containing a list of
-%% tuples: {OriginalFuntion, PreErlang18Function, PostErlang18Function, Arity}
+%% tuples:
+%%
+%% {ErlangVersion, [{OriginalFuntion, PreErlangVersionFunction,
+%%                   PostErlangVersionFunction, Arity}]}
 %%
 %% All these new functions may be exported, and implemented as follows:
 %%
@@ -41,11 +44,11 @@
 %%    code_version:update(?MODULE),
 %%    ?MODULE:OriginalFunction().
 %%
-%% PostErlang18Function() ->
+%% PostErlangVersionFunction() ->
 %%    %% implementation using new time API
 %%    ..
 %%
-%% PreErlang18Function() ->
+%% PreErlangVersionFunction() ->
 %%    %% implementation using fallback solution
 %%    ..
 %%
--- a/src/code_version.erl
+++ b/src/code_version.erl
@@ -32,7 +32,7 @@
 %% The purpose of this functionality is to support the new time API introduced
 %% in ERTS 7.0, while providing compatibility with previous versions.
 %%
-%% `Module` must contain an attribute `version_support` containing a list of
+%% `Module` must contain an attribute `erlang_version_support` containing a list of
 %% tuples: {OriginalFuntion, PreErlang18Function, PostErlang18Function, Arity}
 %%
 %% All these new functions may be exported, and implemented as follows:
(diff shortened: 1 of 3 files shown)


dumbbell:
For me (Erlang 18.2.2), RabbitMQ doesn't start: `time_compat:monotonic_time/0` enters an infinite loop trying to update/reload the module. I'm going to debug this and give more informations.

src/time_compat.erl
```diff
@@ -43,160 +43,242 @@
...
+-version_support(
+   [{monotonic_time, monotonic_time_pre_18, monotonic_time_post_18, 0},
+    {monotonic_time, monotonic_time_pre_18, monotonic_time_post_18, 1},
+    {erlang_system_time, erlang_system_time_pre_18, erlang_system_time_post_18, 0},
+    {erlang_system_time, erlang_system_time_pre_18, erlang_system_time_post_18, 1},
+    {os_system_time, os_system_time_pre_18, os_system_time_post_18, 0},
+    {os_system_time, os_system_time_pre_18, os_system_time_post_18, 1},
+    {time_offset, time_offset_pre_18, time_offset_post_18, 0},
+    {time_offset, time_offset_pre_18, time_offset_post_18, 1},
+    {convert_time_unit, convert_time_unit_pre_18, convert_time_unit_post_18, 0},
```
dumbbell: The infinite loop comes from this line: `convert_time_unit` arity is 3, not 0. Thus, convert_time_unit is not replaced.

src/time_compat.erl
```diff
@@ -43,160 +43,242 @@
...
+monotonic_time_post_18(Unit) ->
     try
-	erlang:monotonic_time()
+        erlang:monotonic_time(Unit)
     catch
-	error:undef ->
-	    %% Use Erlang system time as monotonic time
-	    erlang_system_time_fallback()
+        error:badarg ->
+            erlang:error(badarg, [Unit])
```
dumbbell: Why do you catch exceptions from the official API? I believe the compat modules should not change anything, it should be a transparent wrapper of the API.

src/code_version.erl
```diff
@@ -0,0 +1,234 @@
...
+                false ->
+                    Tree
+            end
+    end.
+
+rename_abstract_functions(ToRename, ToName) ->
+    fun(Tree, {Name, _Arity} = Function) ->
+            case lists:member(Function, ToRename) of
+                true ->
+                    FunctionName = proplists:get_value(Name, ToName),
```
dumbbell: You ignore the arity of the function about to be renamed. `ToName`, returned by `get_name_pairs/2`, looks like:
```erlang
ToName=[{monotonic_time_post_18,monotonic_time}, % Comes from {monotonic_time, monotonic_time_pre_18, monotonic_time_post_18, 0}
        {monotonic_time_post_18,monotonic_time}, % Comes from {monotonic_time, monotonic_time_pre_18, monotonic_time_post_18, 1}
        % ...
```

It doesn't contain the arity information either, but this is important: here you pick the new name from the first matching entry, ie. always the `{monotonic_time, monotonic_time_pre_18, monotonic_time_post_18, 0}` attribute, even if the new name is different for the `/1` function.

src/time_compat.erl
dcorbacho: All those try/catch clauses were in the previous implementation, so I left them there. I can't think of why they could be needed, so if we both agree they can be removed.

dcorbacho: Okay, I found a case where those try/catch do 'something'. I think in the case of `convert_time_unit_fallback` the author wanted to return the exact badarg error the system call would return, that is, with the original function call arguments. In others, like `monotonic_time`, it doesn't change anything. What shall we do @dumbbell ?

michaelklishin:
Merging per discussion in #346 and another round of QA on R16B03 and 18.2.1, with and without HiPE.

PR #47, opened 2016-01-13
Adds handling of exit signals from socket process to rabbit_reader, when terminating channels

Ayanda-D:
Fixes rabbitmq/rabbitmq-server#544.


michaelklishin:
On it.

Ayanda-D:
ok  :+1:

PR #48, opened 2016-01-14
Move random helper to rabbit_misc

binarin:
Needed by https://github.com/rabbitmq/rabbitmq-server/pull/552

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ otp_release:
 # to a tag or a branch. This pleases our git_rmq fetch method in
 # rabbitmq-components.mk and the proper tag/branch is selected in
 # dependencies too.
-before_script: (test "$TRAVIS_TAG" && git checkout "$TRAVIS_TAG") || (test "$TRAVIS_BRANCH" && git checkout "$TRAVIS_BRANCH")
+before_script: (test "$TRAVIS_TAG" && git checkout "$TRAVIS_TAG") || (test "$TRAVIS_BRANCH" && git checkout -b "$TRAVIS_BRANCH")
 
 script: make tests
 
--- a/src/rabbit_misc.erl
+++ b/src/rabbit_misc.erl
@@ -71,6 +71,7 @@
 -export([store_proc_name/1, store_proc_name/2]).
 -export([moving_average/4]).
 -export([get_env/3]).
+-export([random/1]).
 
 %% Horrible macro to use in guards
 -define(IS_BENIGN_EXIT(R),
@@ -259,6 +260,9 @@
 -spec(moving_average/4 :: (float(), float(), float(), float() | 'undefined')
                           -> float()).
 -spec(get_env/3 :: (atom(), atom(), term())  -> term()).
+
+-spec(random/1 :: (non_neg_integer()) -> non_neg_integer()).
+
 -endif.
 
 %%----------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


dumbbell:
Obviously, the change to `.travis.yml` crept in this patch but is unrelated.

Out of curiosity, why did you need to make this change to `.travis.yml`? Did you notice a problem with `git checkout` without `-b`?

binarin:
Because of 2 things:
- Travis build was broken for this PR because it does shallow clone and does detached checkout. But this working copy doesn't contain any information about any branch except 'master'.
- Even if it worked as intended (i.e. travis was doing full checkout), it wouldn't test PR, but only current state of branch where PR is going to be merged to.

And 'git checkout -b $TRAVIS_BRANCH' does the right thing in this case: it just gives the name of target branch to detached head that travis has checked out - so the dependency magic mentioned in comment will work start working.

binarin:
Hm, and now non-PR builds are broken - https://travis-ci.org/rabbitmq/rabbitmq-common/jobs/102444868 (
So the fix need to be more clever.

binarin:
Looks like `-b` needs to be replaced with '-B' - then builds will work both on PR-s and branches.

And so the only thing is left is tags. Because `git checkout "$TRAVIS_TAG"` will actually check out detached head - exactly the situation that it was intented to work around )

dumbbell:
The goal of this checkout is to be on a named reference: the Makefile then uses this name when cloning dependencies. For instance, if the top-level project is on `rabbitmq-server-94`, it will try to use the same branch name in dependencies. If this branch doesn't exist, it will try the "parent" branch (could be `stable`). If this fails again, it falls back to `master`.

Your `git checkout -b` (or even `-B` as you suggest) is interesting because, `git checkout` alone does not work (ie. does not create a local branch based on the remote branch with the same name) when there are multiple remotes configured, which must be the case for pull requests.

Then for tags, it should be ok as is, because I don't see how we can end up in Travis with multiple remotes and the need for a tag which doesn't exist in the "upstream" repository (eg. rabbitmq/rabbitmq-common).

I believe your patch (with `-B`) is correct, it handles all situations. Do you see a corner case which would break?

binarin:
Primary source of the problems is that travis does only partial checkout:
- For PR there will be only default branch and no tags.
- For branch build there will be only that branch and no tags.
- I haven't checked for tags, but I believe workdir will contain just that tag and default branch.

To make it work as you described:
- For PRs best name we could give to checked out version is `git checkout -B $TRAVIS_BRANCH`', so it will be named as a target branch. We also need to fetch stable/master branches so "parent" detection would work.
- For branches we could do just `git checkout $TRAVIS_BRANCH`, because in this case branch is actually fetched from remote. And "parent" detection will still need additional fetch of stable/master branches. `-B` could be added to options, it'll be just a no-op.
- For tags `git checkout $TRAVIS_TAG` will result in detached head situation. Two possibilities here:
  - Use `git checkout -B $TRAVIS_TAG` to  create branch that has same name as the currently being built tag
  - Use `git describe --tag` somewhere inside all this makefile magic

So the best effort to conform to your description (without delving into makefile magic) will look like:
```
git fetch origin master
git fetch origin stable
BRANCH_TO_CREATE=${TRAVIS_TAG:-$TRAVIS_BRANCH}
git checkout -B "${BRANCH_TO_CREATE}"
```

dumbbell:
@binarin: Hi! FYi, I made the following change to `.travis.yml` in `rabbit-common`, based on your suggestion:
https://github.com/rabbitmq/rabbitmq-common/blob/master/.travis.yml#L31-L33

It seems to work for tags and push to branches. It won't work for pull request (or in a degraded way) because Travis doesn't provide the source branch name. I don't see a solution to this issue, but it's acceptable.

Thank you for your help!

PR #49, opened 2016-01-25
Update channel interceptors on register/unregister

hairyhum:
Fixes https://github.com/rabbitmq/rabbitmq-server/issues/559
Callbacks to update interceptors.
Related to https://github.com/rabbitmq/rabbitmq-server/pull/578

Callback cause all channels to refresh interceptors list.

### DIFF
--- a/src/rabbit_channel.erl
+++ b/src/rabbit_channel.erl
@@ -150,6 +150,8 @@
 
 -define(MAX_PERMISSION_CACHE_SIZE, 12).
 
+-define(REFRESH_TIMEOUT, 15000).
+
 -define(STATISTICS_KEYS,
         [pid,
          transactional,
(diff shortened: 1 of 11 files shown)


videlalvaro:
This PR seems fine but see my comment on #578

src/rabbit_channel.erl
```diff
@@ -337,6 +338,12 @@ refresh_config_local() ->
       list_local()),
     ok.
 
+refresh_interceptors() ->
+    rabbit_misc:upmap(
+      fun (C) -> gen_server2:call(C, refresh_interceptors, infinity) end,
```
michaelklishin: Please use a finite timeout here, e.g. 15 seconds.

PR #51, opened 2016-02-08
Optimisation

dcorbacho:
* Use lists:keyfind instead of proplists to speed up the management stats
Derives from https://github.com/rabbitmq/rabbitmq-management/issues/41

### DIFF
--- a/src/rabbit_misc.erl
+++ b/src/rabbit_misc.erl
@@ -882,8 +882,20 @@ is_process_alive(Pid) ->
     lists:member(Node, [node() | nodes()]) andalso
         rpc:call(Node, erlang, is_process_alive, [Pid]) =:= true.
 
-pget(K, P) -> proplists:get_value(K, P).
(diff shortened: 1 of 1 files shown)


michaelklishin:
@dcorbacho I'm sure this was a change driven by benchmarks used in https://github.com/rabbitmq/rabbitmq-management/issues/41 but out of curiosity, how much difference does it make for this specific function, %-wise?

dcorbacho:
@michaelklishin proplists is ~7x slower than lists. Querying the queues in management DB, for 62k queues it was called at least 824k times, so the improvement is noticeable.

Lists are a BIF, while proplists is not and does some verification on the structure of the list: https://github.com/erlang/otp/blob/maint/lib/stdlib/src/proplists.erl#L208

michaelklishin:
Thanks for the clarification.

michaelklishin:
This can be merged separately from the rest of `management#41`.

PR #52, opened 2016-02-09
Prevent infinite loop on 'queue.declare'

binarin:
`rabbit_channel:handle_method` for `queue.declare` could go into infinite loop, when it tries do define a queue which got somehow corrupted. This fix at least will help with symptoms of this corruption - channel will quickly become unstuck.

Loop happens due to somewhat special handling of `not_found` in 'queue.declare' - it is treated not as an error here, but as a permission to proceed with queue creation. As the base case of recursion in `rabbit_amqqueue:with/4` is actuallly about some unrecoverable error, it makes sense to always return `absent` error here. It will still result in `not_found` error for clients, but internally it'll be handled correctly.

### DIFF
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -413,10 +413,12 @@ not_found_or_absent_dirty(Name) ->
 with(Name, F, E) ->
     with(Name, F, E, 2000).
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you.

PR #54, opened 2016-02-12
Remove custom stderr formatting

### LINKED ISSUE #53: Segfault while printing errors to stderr with erlang 18.X
Looks like code at https://github.com/rabbitmq/rabbitmq-common/blob/b6c0fa46446181ce84b8856f12e50f33481065ff/src/rabbit_misc.erl#L657
is incomptabile with erlang 18.X and its bugs. It causes every other unsuccessful `rabbitmqctl` run to segfault and dump core.

Erlang 18.X VM is really unhappy with things like:
```
./bin/erl -noshell -eval 'Port = open_port({fd, 0, 2}, [out]), port_command(Port, "a"), port_close(Port), erlang:halt(10)'
```

And I think OTP team don't currentl have solution themselves, because they use workarounds like https://github.com/erlang/otp/commit/ba531a173d0dc6906f8492f3bfa0b2cd40e2951a :smile: 

So I think it's better to remove this troublesome piece of code from rabbitmq, because nobody knows then the erlang 18.X itself will be fixed. And because ops use this segfault as an excuse to stop digging into some other problems further :smile:

binarin:
Looks like all issues were fixed in erlang 17.0. After 1000 thousand invocations of rabbitmqctl on busy machine nothing was lost neither on linux
```
for i in $(seq 1 1000); do ./scripts/rabbitmqctl status 2>&1; done | grep 'cookie hash' | wc -l 
```
nor on windows
```
for /l %x in (1, 1, 1000) do rabbitmqctl.bat status 2>&1 | findstr /c:"cookie hash" >> cookie-hash.txt
```

Opening several ports for single fd is considered undefined behaviour in
erlang. It's safe to replace this whole function with `io:format` when
erlang 17 or later is used. Because writing to standard_error with
io:format is synchronous - after this call has returned data was
definitely sent to the port. And `erlang:halt/` guarantees that this
data will be flushed afterwards.

Closes #53

### DIFF
--- a/src/rabbit_misc.erl
+++ b/src/rabbit_misc.erl
@@ -655,18 +655,7 @@ format_many(List) ->
     lists:flatten([io_lib:format(F ++ "~n", A) || {F, A} <- List]).
 
 format_stderr(Fmt, Args) ->
-    case os:type() of
-        {unix, _} ->
-            Port = open_port({fd, 0, 2}, [out]),
-            port_command(Port, io_lib:format(Fmt, Args)),
-            port_close(Port);
-        {win32, _} ->
-            %% stderr on Windows is buffered and I can't figure out a
-            %% way to trigger a fflush(stderr) in Erlang. So rather
-            %% than risk losing output we write to stdout instead,
-            %% which appears to be unbuffered.
-            io:format(Fmt, Args)
-    end,
+    io:format(standard_error, Fmt, Args),
     ok.
 
 unfold(Fun, Init) ->


michaelklishin:
@binarin thanks. Should we do anything for `stable` then, which has to support `R16B03` at the moment?

binarin:
I've looked into R16B03 sources and `io:format` to `standard_error` is also synchronous there. So maybe this patch could be applied as is to `stable`. But I'm not 100% sure about windows part.

PR #56, opened 2016-02-18
Adds channel_operation_timeout when channels notify queues of shutdown

Ayanda-D:
Fixes rabbitmq/rabbitmq-server#248.

### DIFF
--- a/src/rabbit_misc.erl
+++ b/src/rabbit_misc.erl
@@ -71,6 +71,7 @@
 -export([store_proc_name/1, store_proc_name/2]).
 -export([moving_average/4]).
 -export([get_env/3]).
+-export([get_channel_operation_timeout/0]).
 -export([random/1]).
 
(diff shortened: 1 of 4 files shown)


src/rabbit_misc.erl
```diff
@@ -71,6 +71,7 @@
 -export([store_proc_name/1, store_proc_name/2]).
 -export([moving_average/4]).
 -export([get_env/3]).
+-export([get_channel_operation_timeout/0]).
```
hairyhum: Shouldn't this function be in `rabbit_channel`, since it defines a channel setting?

Ayanda-D: @hairyhum yes it could. `channel_operation_timeout` is going to be used in various places, and had a brief chat with @michaelklishin about making it accessible from a common module. & chose to use `rabbit_misc`.

PR #58, opened 2016-02-22
rabbitmq_versions key for plugin

hairyhum:
Related to https://github.com/rabbitmq/rabbitmq-server/pull/642

https://github.com/rabbitmq/rabbitmq-server/issues/591

### DIFF
--- a/include/rabbit.hrl
+++ b/include/rabbit.hrl
@@ -126,11 +126,13 @@
                  type,             %% 'ez' or 'dir'
                  dependencies,     %% [{atom(), string()}]
                  location,         %% string()
-                 % List of supported broker versions
+                 %% List of supported broker version ranges,
+                 %% e.g. ["3.5.7", "3.6.1"]
                  broker_version_requirements, %% [string()]
-                 % Proplist of supported dependencies versions
-                 % E.g. [{dep_name, ["3.5.1", "3.6.0"]}, 
-                 %       {other_dep_name, ["3.6.0"]}]
+                 %% Proplist of supported dependency versions,
+                 %% e.g. [{rabbitmq_management, ["3.5.7", "3.6.1"]}, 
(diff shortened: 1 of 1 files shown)


include/rabbit.hrl
```diff
@@ -120,12 +120,15 @@
...
-                 dependencies,  %% [{atom(), string()}]
-                 location}).    %% string()
+-record(plugin, {name,             %% atom()
+                 version,          %% string()
+                 description,      %% string()
+                 type,             %% 'ez' or 'dir'
+                 dependencies,     %% [{atom(), string()}]
+                 location,         %% string()
+                 plugins_versions, %% [{atom(), string()}]
+                 rabbitmq_versions %% [string()]
```
michaelklishin: This field does make sense to me but I'd recommend renaming it to `supported_server_versions`.

include/rabbit.hrl
```diff
@@ -120,12 +120,15 @@
...
-                 type,          %% 'ez' or 'dir'
-                 dependencies,  %% [{atom(), string()}]
-                 location}).    %% string()
+-record(plugin, {name,             %% atom()
+                 version,          %% string()
+                 description,      %% string()
+                 type,             %% 'ez' or 'dir'
+                 dependencies,     %% [{atom(), string()}]
+                 location,         %% string()
+                 plugins_versions, %% [{atom(), string()}]
```
michaelklishin: Per discussion on Slack, lets rename this to `dependency_version_requirements `.

michaelklishin: I'd also use an example in the comments. Type spec-ish comments are useful but so are specific examples.

include/rabbit.hrl
michaelklishin: I'd also use an example in the comments. Type spec-ish comments are useful but so are specific examples.

PR #59, opened 2016-02-22
Ignore result from notify_queues/1

Ayanda-D:
Ref: rabbitmq/rabbitmq-server#248


michaelklishin:
Thank you, @Ayanda-D. My bad.

PR #62, opened 2016-02-26
SPIKE rabbit_reader state machine(s) refactoring

hairyhum:
Idea to refactor blocking states in `rabbit_reader`
Currently there are `running` `blocking` and `blocked` states and `throttle` field, which contain state information. So transitions indirectly depend on this state information.

So proposal is to change this state machine: https://github.com/hairyhum/pics/blob/master/old.png
Into this: https://github.com/hairyhum/pics/blob/master/new.png

So there will be more sane transitions which will not depend so much on throttle values during transition.

See comments [here](https://github.com/rabbitmq/rabbitmq-server/issues/499#issuecomment-189233764)

### DIFF
--- a/src/rabbit_reader.erl
+++ b/src/rabbit_reader.erl
@@ -151,11 +151,12 @@
   %% a set of the reasons why we are
   %% blocked: flow, resource
   block_reasons,
-  % map()
+  % a map of blocking reason => connection.blocked message
   block_messages,
-  %% boolean() true if received any publishes
+  %% true if received any publishes, false otherwise
   can_block,
-  % boolean()
+  %% true if we had we sent a connection.blocked,
+  %% false otherwise
   connection_blocked_message_sent
 }).
 
--- a/src/rabbit_reader.erl
+++ b/src/rabbit_reader.erl
@@ -380,6 +380,7 @@ start_connection(Parent, HelperSup, Deb, Sock) ->
                 throttle            = #throttle{
                                          alarmed_by      = [],
                                          last_blocked_at = never,
+                                         have_publishes  = false,
                                          blocked_reasons = sets:new(),
                                          blocked_messages = maps:new(),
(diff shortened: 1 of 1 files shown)


hairyhum:
After some analysis of code and SM there are following properties:
`had_publish` - there were publishes
`cf_blocked` - process blocked by credit_flow
`other_blocked` - process blocked by not credit flow, so it should send `connection.block`

Current `blocked` state should be reached if something blocking connection and there were publishes.
`connection_blocked = had_publish ^ (cf_blocked v other_blocked)`

Sending `connection.block` should be performed when connection is being blocked and it's not blocked by credit flow:

`connection_block_send = connection_blocked ^ ¬cf_blocked = had_publish ^ other_blocked`
 
In connection lifetime only `connection_blocked` state is make sense. In blocking scope there can be `blocking_message_sent` and `blocking_message_not_sent` states. Everything else can be variables, which define (with formulas mentioned above) to change this two state machines (each of them is of two states).

michaelklishin:
There are two reasons why a connection process may be blocked:

 * Credit flow
 * Resource alarms

So "other" can be a lot more specific. We also don't want to send `connection.blocked` more than once
until it is unblocked.

hairyhum:
#499 requires another reason for connection to be blocked, that's the point of all this work, actually.
And state `blocking_message_sent` needed to avoid sending `connection.blocked` multiple times.

michaelklishin:
@hairyhum `blocking_message_sent` should probably be `connection_blocked_message_sent`. The method is `connection.blocked`, not `connection.blocking`.

michaelklishin:
@hairyhum ok, good point about #499 introducing another blocking reason.

hairyhum:
Tests are passing. Not sure about naming though.

michaelklishin:
This was squared and merged into master manually.

PR #66, opened 2016-03-02
Move types and auth_backend modules over from rabbitmq-server.

bdshroyer:
* Remove modules from rabbitmq-server and add them to rabbitmq-common.

Addresses [#65](https://github.com/rabbitmq/rabbitmq-common/issues/65).


bdshroyer:
Should be merged into stable, not master.

PR #71, opened 2016-03-10
Support for parallel info emission

binarin:
Complement for https://github.com/rabbitmq/rabbitmq-server/pull/683

- `rabbit_control_misc:wait_for_info_messages/5` can wait for results of
  more than one emitting map.
- Stop passing arround InfoItemKeys in
  `rabbit_control_misc:wait_for_info_messages/5`, the same information
  could be directly encoded in DisplayFun closure.
- Add `emit` to function names, to avoid confusion with regular ones
  which return result directly.

### DIFF
--- a/src/rabbit_control_misc.erl
+++ b/src/rabbit_control_misc.erl
@@ -17,7 +17,8 @@
 -module(rabbit_control_misc).
 
 -export([emitting_map/4, emitting_map/5, emitting_map_with_exit_handler/4,
-         emitting_map_with_exit_handler/5, wait_for_info_messages/5,
+         emitting_map_with_exit_handler/5, wait_for_info_messages/6,
+         spawn_emitter_caller/7, await_emitters_termination/1,
          print_cmd_result/2]).
 
 -ifdef(use_specs).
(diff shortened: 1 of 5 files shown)


src/rabbit_control_misc.erl
```diff
@@ -69,25 +69,27 @@ step_with_exit_handler(AggregatorPid, Ref, Fun, Item) ->
             ok
     end.
 
-wait_for_info_messages(Pid, Ref, ArgAtoms, DisplayFun, Timeout) ->
-    _ = notify_if_timeout(Pid, Ref, Timeout),
-    wait_for_info_messages(Ref, ArgAtoms, DisplayFun).
+wait_for_info_messages(Pid, Ref, DisplayFun, Timeout, Chunks) ->
```
dcorbacho: I'd rename `Chunks` to `NodesCount` and `NodesLeft` or something like that, so that it is obvious what exactly the function is waiting for.

michaelklishin:
@binarin I have conflicts when merging this.

binarin:
Updated patch.

PR #74, opened 2016-03-18
code_version: Acquire a lock to protect concurrent reloads

### LINKED ISSUE #73: `code_version` fails to handle concurrent load of modules
When several parallel calls are made to `time_compat` or `ssl_compat` before they were reloaded by `code_version`, the code server returns an error and `code_version` throws an exception.

Here is the message from the code server:
```
=ERROR REPORT==== 18-Mar-2016::11:00:20 ===
Module time_compat must be purged before loading
```

References rabbitmq/rabbitmq-server#346, rabbitmq/rabbitmq-server#347.

dumbbell:
Before this patch, concurrent calls to `code_version` for the same module would lead to one call succeeding and the other calls getting an exception because the code server complains about a non-purged copy.

Now:

1. We acquire a lock for the given module name
2. We check the filename of the already loaded module to replace: if it matches the fake filename used by the reload, it means the module was replaced in a concurrent call. The next step is skipped.
3. We reload the module.
4. We release the lock.

Fixes #73.

### DIFF
--- a/src/code_version.erl
+++ b/src/code_version.erl
@@ -67,13 +67,21 @@ update(Module) ->
 %% Internal functions
 %%----------------------------------------------------------------------------
 load_code(Module, Code) ->
-    unload(Module),
-    case code:load_binary(Module, "loaded by rabbit_common", Code) of
-        {module, _} ->
+    LockId = {{?MODULE, Module}, self()},
+    FakeFilename = "Loaded by rabbit_common",
+    global:set_lock(LockId, [node()]),
+    case code:which(Module) of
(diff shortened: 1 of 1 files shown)


michaelklishin:
Do we really need to use `global` here? This is node-local state.

dumbbell:
I don't know any other locking API in Erlang. The call to `global:set_lock/2` specifies the lock is valid only on the local node.

Do you suggest a better way?

michaelklishin:
Ah, I've missed the local node argument. Sorry about the noise.

PR #76, opened 2016-03-23
Fix `rabbitmqctl list_consumers`

binarin:
It makes no sense to concatenate information about several consumers
into single proplist.

`rabbit_amqqueue:consumers_all/1` is already doing `lists:append/1`, so
it is already capable of handling zero or more consumers reported.

And for `rabbitmqctl list_consumers` we need to support printing
lists-of-lists, for the same purpose of reporting zero or more consumers.

Another part for https://github.com/rabbitmq/rabbitmq-server/issues/701


michaelklishin:
This was submitted against `master`. Please re-submit.

PR #82, opened 2016-04-05
Save listener socket options in Mnesia (DO NOT MERGE)

essen:
Work in progress. This adds a field to Mnesia; we need to give a proper upgrade path.

Part of https://github.com/rabbitmq/rabbitmq-management/issues/163


michaelklishin:
@essen there are [upgrade functions](https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_upgrade_functions.erl) but because this changes database schema this cannot go into `stable`.

PR #84, opened 2016-04-05
Capture detailed Erlang logs on connection failure

dcorbacho:
Closes https://github.com/rabbitmq/rabbitmq-server/issues/401

### DIFF
--- a/src/rabbit_error_logger_handler.erl
+++ b/src/rabbit_error_logger_handler.erl
@@ -158,7 +158,7 @@ code_change(_OldVsn, State, _Extra) ->
 format({check_dflag_xnc_failed, _What}) ->
     {"  * Remote node uses an incompatible Erlang version ~n", []};
 format({recv_challenge_failed, no_node, Node}) ->
-    {"  * Hostname mismatch ~p~n", [Node]};
+    {"  * Hostname mismatch: node ~p believes its host is different. Please ensure that hostnames resolve the same way locally and on ~p~n", [Node, Node]};
 format({recv_challenge_failed, Error}) ->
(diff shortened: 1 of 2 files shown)


src/rabbit_error_logger_handler.erl
```diff
@@ -0,0 +1,175 @@
...
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+format({check_dflag_xnc_failed, _What}) ->
+    {"  * Remote node uses an incompatible Erlang version ~n", []};
+format({recv_challenge_failed, no_node, Node}) ->
+    {"  * Hostname mismatch ~p~n", [Node]};
```
michaelklishin: I feel we can improve this further. @dcorbacho this case is when remote node name as seen by the remote node itself differs from what the local node used, is my understanding correct? If so, what does the `Node` value indicate?

michaelklishin:
I've updated the wording, @dcorbacho can you please verify that my understanding of each case is correct?

michaelklishin:
Also, this depends on an OTP patch. Is it included in a released Erlang version?

src/rabbit_error_logger_handler.erl
dcorbacho: `Node` is the remote node name as seen by the remote node itself. The name is received in the challenge exchange, and it is the local node that rejects it.

dcorbacho:
@michaelklishin your wording is much better than mine, all correct.

dcorbacho:
The OTP patch is not in any released Erlang version yet. It is merged into master, so it will be available in the next major release (estimated Q2 2016).

PR #105, opened 2016-06-09
Removed wrong included file.

kunalsaxena:
- removed include of unavailable file **rabbit_framing.hrl** in **rabbit_misc.erl**

- Getting dependency error with rebar because of this.

### DIFF
--- a/src/rabbit_misc.erl
+++ b/src/rabbit_misc.erl
@@ -16,7 +16,6 @@
 
 -module(rabbit_misc).
 -include("rabbit.hrl").
--include("rabbit_framing.hrl").
 -include("rabbit_misc.hrl").
 
 -export([method_record_type/1, polite_pause/0, polite_pause/1]).


michaelklishin:
That file is generated. If you run `make` in this repo, you will find it under `include`.

michaelklishin:
Now, it's a different thing if this file isn't actually needed in `rabbit_misc`. But claiming it to be "unavailable" is wrong and means you don't understand how this project is built. While we would be interested in making it easier to use with Rebar, it's a matter of Rebar [in]flexibility to make it possible to build projects such as this one.

kunalsaxena:
I was including this as rebar dependency into ejabberd and was getting error because **rabbit_framing.hrl** was not there.
I am adding rabbitmq-common and amqp-client rebar dependencies as 

```
{rabbit_common, ".*", {git, "https://github.com/rabbitmq/rabbitmq-common"}},
        {amqp_client, ".*", {git, "git://github.com/rabbitmq/rabbitmq-erlang-client.git"}},

``` 
I am getting following error while compiling ejabberd

```
==> rabbit_common (compile)
Compiled src/rabbit_authn_backend.erl
/home/kunal/Dev/Project/Snatch/snatch_ejabberd_backend/deps/rabbit_common/src/rabbit_misc.erl:19: can't find include file "rabbit_framing.hrl"
ERROR: compile failed while processing /home/kunal/Dev/Project/Snatch/snatch_ejabberd_backend/deps/rabbit_common: rebar_abort
make: *** [deps/.built] Error 1
```

I am not building rabbitmq-common separately and using this as dependency. I may have miss something so please help correcting this error.

binarin:
Erlang client from official repository is not rebar-compatible, you should use compatible version at https://github.com/jbrisbin/amqp_client

Also, you don't need explicit dependency on `rabbit_common`.

Given that `rebar3` is an official build tool, maybe the time has come to implement rebar-compatibily in  `rabbitmq-erlang-client` and `rabbitmq-common` :smile:

binarin:
Or even better, just use dependency from hex.pm. Its as easy as just referencing `amqp_client` without any details:
`{deps, [amqp_client]}.`
You need patched amqp_client from previous comment only for rebar2

michaelklishin:
@binarin please file a new issue for that?

binarin:
That Issue already exists - https://github.com/rabbitmq/rabbitmq-erlang-client/issues/35

michaelklishin:
👍

PR #107, opened 2016-06-16
Report reductions and garbage collection options in connection stats

hairyhum:
Part of https://github.com/rabbitmq/rabbitmq-management/issues/211


hairyhum:
Added https://github.com/rabbitmq/rabbitmq-management/issues/210

PR #111, opened 2016-06-26
Compile on OTP 19.0

michaelklishin:
Part of https://github.com/rabbitmq/rabbitmq-server/issues/860.

There are two primary changes:

 * Type spec syntax (backwards compatible)
 * `rand` is used instead of `random`


michaelklishin:
OK, apparently the `rand` module is only [available in `18.0+`](https://github.com/erlang/otp/commit/95aff702b5e4b21ec277b1e0125f639ce30f997a), so this needs more work.

michaelklishin:
@dumbbell thank you, this now needs merging into master.

dumbbell:
This PR is merged to master.

binarin:
Now it fails to build on older erlangs (i.e. 18.3.4) with
```
src/pmon.erl:43: referring to built-in type dict as a remote type; please take out the module name
```

binarin:
Ah, sorry, I was accidentally using R16B03

PR #115, opened 2016-06-30
Don't spew stacktrace on join when already member

binarin:
Example of stacktrace - https://github.com/rabbitmq/rabbitmq-server/issues/709


michaelklishin:
I'd use "a member of this cluster" instead of "Properly clustered."

Otherwise 👍.

binarin:
Done.

michaelklishin:
Thank you. This will go into `3.6.4` as the last RC we have planned for `3.6.3` has shipped just a couple of hours ago :(

PR #118, opened 2016-07-14
Make `rabbit_health_check` completely node-local

binarin:
Part of https://github.com/rabbitmq/rabbitmq-server/issues/818

Also:
- fix queues health-checking - `rabbit_amqqueue:info_all` was expecting
  vhost argument, but was passed `[pid]` instead. So no health check was
  actually performed. Proper way is to iterate over all vhosts.
- delegate all error/timeout handling to `rabbit_cli`, so no more magic
  `70` and `68` and no more `case`-ing on `badrpc` that is already
  present in other places.

### DIFF
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -25,7 +25,7 @@
          check_exclusive_access/2, with_exclusive_access_or_die/3,
          stat/1, deliver/2, requeue/3, ack/3, reject/4]).
 -export([list/0, list/1, info_keys/0, info/1, info/2, info_all/1, info_all/2,
-         info_all/6]).
+         info_all/6, info_local/1]).
 -export([list_down/1]).
 -export([force_event_refresh/1, notify_policy_changed/1]).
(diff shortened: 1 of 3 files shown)


michaelklishin:
@binarin hm, I don't see `rabbit_cli` referenced. It's a good idea but we should keep in mind that `rabbit_control_main` and friends are going away in master in a couple of months. All CLI tools will be in a separate repo.

michaelklishin:
OK, it's in https://github.com/rabbitmq/rabbitmq-server/pull/883, sorry.

binarin:
Delegation to `rabbit_cli` is implicit. Before this patch, `rabbit_health_check` was transforming different `{badrpc, _}` (and other nasty things) to errors and exit-codes by itself. But `rabbit_cli` has big case statement which does the same. And I think it's better to reuse it.

michaelklishin:
```
action(node_health_check, Node, _Args, _Opts, Inform)
```

uses `rabbit_health_check:node/1` but only `rabbit_health_check:node/2` is exported, therefore
things fail with an `undef`.

michaelklishin:
Disregard my comment above, I had a bad merge locally.

PR #120, opened 2016-07-15
Use user specified connection name in logs

binarin:
Remake of #101 - now connection name in managment UI and HTTP UI is untouched.

In addition to https://github.com/rabbitmq/rabbitmq-server/issues/104


michaelklishin:
Thank you!

michaelklishin:
I've noticed we don't use the augmented name in channel errors but that's OK: it should already be quite a bit easier to track down an error to a client/app with this.

PR #121, opened 2016-07-22
Keep track of connections, introduce per-vhost limits

michaelklishin:
Part of rabbitmq/rabbitmq-server#500 and rabbitmq/rabbitmq-server#627.
Lays ground work for rabbitmq/rabbitmq-server#607.

### DIFF
--- a/src/rabbit_ct_broker_helpers.erl
+++ b/src/rabbit_ct_broker_helpers.erl
@@ -777,26 +777,26 @@ format_ipaddr_for_uri(
 add_vhost(Config, VHost) ->
     add_vhost(Config, 0, VHost).
 
-add_vhost(Config, NodeIndex, VHost) ->
-    rabbit_ct_broker_helpers:rpc(Config, NodeIndex, rabbit_vhost, add, [VHost]).
+add_vhost(Config, Node, VHost) ->
+    rabbit_ct_broker_helpers:rpc(Config, Node, rabbit_vhost, add, [VHost]).
 
 delete_vhost(Config, VHost) ->
     delete_vhost(Config, 0, VHost).
 
-delete_vhost(Config, NodeIndex, VHost) ->
-    rabbit_ct_broker_helpers:rpc(Config, NodeIndex, rabbit_vhost, delete, [VHost]).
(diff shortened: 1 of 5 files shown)


src/rabbit_ct_broker_helpers.erl
```diff
@@ -719,6 +726,25 @@ format_ipaddr_for_uri(
...
     "[" ++ Res1 ++ "]".
 
+
+%% Virtual host management
+
+add_vhost(Config, VHost) ->
+    rabbit_ct_broker_helpers:rpc(Config, 0, rabbit_vhost, add, [VHost]).
+
+delete_vhost(Config, VHost) ->
+    rabbit_ct_broker_helpers:rpc(Config, 0, rabbit_vhost, delete, [VHost]).
```
dumbbell: Could you please not hard-code the node and take it as an argument instead?

src/rabbit_ct_broker_helpers.erl
```diff
@@ -719,6 +726,25 @@ format_ipaddr_for_uri(
...
+
+set_full_permissions(Config, VHost) ->
+    set_permissions(Config, <<"guest">>, VHost, <<".*">>, <<".*">>, <<".*">>).
+set_full_permissions(Config, Username, VHost) ->
+    set_permissions(Config, Username, VHost, <<".*">>, <<".*">>, <<".*">>).
+set_permissions(Config, Username, VHost, ConfigurePerm, WritePerm, ReadPerm) ->
+    rabbit_ct_broker_helpers:rpc(Config, 0,
```
dumbbell: Same comment as above.

src/rabbit_ct_broker_helpers.erl
michaelklishin: I can introduce a new function head but since vhost and user operations are cluster-wide, would they be used often?

dumbbell: Nodes may not be clustered. As this is the responsability of the testcase, I would indicate the node in the test, not in `rabbit_ct_broker_helpers.erl`.

PR #129, opened 2016-08-26
Remove it. Use iolist_to_binary instead.

lemenkov:
This looks like half-implemented stub for ensuring that lists/binaries will be represented as binaries. No one is using it, it lacks specs (so I'm unsure what was the plan actually), and there are better alternatives (iolist_to_binary). So let's remove it, thus saving space.


michaelklishin:
MQTT and tracing plugins use it.

lemenkov:
GitHub search sucks.

michaelklishin:
Both IDEA and grep over `deps/*` revealed it quite quickly ;)

PR #134, opened 2016-09-19
Provide a way to limit max number of queues in a vhost

michaelklishin:
Part of https://github.com/rabbitmq/rabbitmq-server/issues/501.


src/rabbit_channel.erl
```diff
@@ -1324,6 +1333,7 @@ handle_method(#'queue.declare'{queue       = QueueNameBin,
                     end,
     QueueName = rabbit_misc:r(VHostPath, queue, ActualNameBin),
     check_configure_permitted(QueueName, State),
+    check_vhost_queue_limit(QueueName, VHostPath),
```
kjnilsson: Because the check is done before checking if the queue already exists it will fail if a user tries to declare an already existing queue after the limit has been reached.

michaelklishin: Good point.

PR #135, opened 2016-09-26
Introduce rabbit_peer_discovery_backend

michaelklishin:
A new behaviour for pluggable cluster peer discovery.

Borrowed from [1] with a bit of renaming and a blessing from Gavin M. Roy.

1. https://github.com/aweber/rabbitmq-autocluster/blob/master/src/autocluster_backend.erl

Part of rabbitmq/rabbitmq-server/issues/486.

### DIFF
--- a/src/rabbit_peer_discovery_backend.erl
+++ b/src/rabbit_peer_discovery_backend.erl
@@ -28,7 +28,7 @@
 %%
 %% The Initial Developer of the Original Code is AWeber Communications.
 %% Copyright (c) 2014-2015 AWeber Communications
-%% Copyright (c) 2016-Current Pivotal Software, Inc.  All rights reserved.
+%% Copyright (c) 2016 Pivotal Software, Inc.  All rights reserved.
 %%
 
 -module(rabbit_peer_discovery_backend).
--- a/src/rabbit_peer_discovery_backend.erl
+++ b/src/rabbit_peer_discovery_backend.erl
@@ -1,0 +1,42 @@
+%% Copyright (c) 2014-2015 AWeber Communications
+%% All rights reserved.
+%%
+%% Redistribution and use in source and binary forms, with or without modification,
+%% are permitted provided that the following conditions are met:
+%%
+%%  * Redistributions of source code must retain the above copyright notice, this
+%%    list of conditions and the following disclaimer.
+%%  * Redistributions in binary form must reproduce the above copyright notice,
+%%    this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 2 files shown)


michaelklishin:
Submitted primarily so that we can discuss license headers with legal.

src/rabbit_peer_discovery_backend.erl
```diff
@@ -0,0 +1,44 @@
...
+%% DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+%% LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
+%% OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+%% ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+%%
+%% The Original Code is rabbitmq-autocluster.
+%%
+%% The Initial Developer of the Original Code is AWeber Communications.
+%% Copyright (c) 2014-2015 AWeber Communications
+%% Copyright (c) 2016-Current Pivotal Software, Inc.  All rights reserved.
```
dumbbell: There is a copyright line at the beginning of this file too. I would put all of them in a single place.

Also, I'm pretty sure "2016-Current" is forbidden. It should be "2016" only.

src/rabbit_types.erl
```diff
@@ -151,6 +151,8 @@
 
 -type(protocol_name() :: 'amqp0_8' | 'amqp0_9_1' | 'amqp1_0' | 'mqtt' | 'stomp' | any()).
 
+-type(node_type() :: 'disc' | 'disk' | 'ram').
```
dumbbell: Why do we need to support two spellings of the word "disk"? I find it confusing, and it doesn't help pattern matching. If we want to support both where a user can type this, then the input should be converted to a single form immediately.

michaelklishin:
Because there are two different spellings used throughout the English speaking world.

dumbbell:
I understand that. My point is why do we need to support both internally? It's only needed in user interfaces.

michaelklishin:
We don't but it will require more diligence from us and plugin developers. But I don't have a strong opinion on this.

kjnilsson:
Would `disk` even work? - looking at this code [here])https://github.com/rabbitmq/rabbitmq-server/pull/989/files#diff-c8e629f12e7b6a73d53dbde4e57d2df5R126) I'm not sure it would.

PR #138, opened 2016-10-12
Broaden version format support

camelpunch:
We need to support prerelease suffix, to use for nightlies and release candidates.
Includes special case for 3.6.6, which is a breaking 'patch'. 💔

Uses https://github.com/erlware/erlware_commons - see commit message for justification.

### DIFF
--- a/src/ec_semver.erl
+++ b/src/ec_semver.erl
@@ -326,6 +326,12 @@ internal_pes(Vsn, LVsn) ->
 eql_test() ->
     ?assertMatch(true, eql("1.0.0-alpha",
                            "1.0.0-alpha")),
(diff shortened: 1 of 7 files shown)


src/ec_semver.erl
```diff
@@ -0,0 +1,717 @@
...
+              andalso
+                (AlphaA > AlphaB))))
+     orelse
+       ((MMPA =:= MMPB)
+        andalso
+          (AlphaA =:= AlphaB)
+        andalso
+          ((PatchB =:= [] andalso PatchA =/= [])
+           orelse
+           PatchA > PatchB))).
```
camelpunch: `ec_semver.erl` is imported code, so a change would probably warrant a PR upstream.

PR #141, opened 2016-10-28
Create .ez for multi-app plugins

binarin:
Without this change applications in `apps/` were compiled, but not
bundled into `.ez` files.


michaelklishin:
N.B.: some future work on `rabbitmq-autocluster` seems to depend on this.

PR #145, opened 2016-11-17
Add rabbit_data_coercion:to_integer/1

acogoluegnes:
Part of rabbitmq/rabbitmq-server#1030

### DIFF
--- a/src/rabbit_data_coercion.erl
+++ b/src/rabbit_data_coercion.erl
@@ -18,16 +18,20 @@
 
 -export([to_binary/1, to_list/1, to_atom/1, to_integer/1]).
 
+-spec to_binary(Val :: binary() | list()) -> binary().
 to_binary(Val) when is_list(Val) -> list_to_binary(Val);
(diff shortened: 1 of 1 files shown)


src/rabbit_data_coercion.erl
```diff
@@ -27,3 +27,7 @@ to_list(Val) when is_binary(Val) -> binary_to_list(Val).
 to_atom(Val) when is_atom(Val)   -> Val;
 to_atom(Val) when is_list(Val)   -> list_to_atom(Val);
 to_atom(Val) when is_binary(Val) -> binary_to_atom(Val, utf8).
+
+to_integer(Val) when is_integer(Val) -> Val;
+to_integer(Val) when is_list(Val)    -> list_to_integer(Val);
+to_integer(Val) when is_binary(Val)  -> binary_to_list(Val).
```
michaelklishin: Looks like a missing call to `list_to_integer/1`.

PR #149, opened 2016-11-24
Add dependencies on compiler and syntax_tools

essen:
While they don't need to be started, we do depend on them,
so should list them as dependencies. This will ensure that
they will be included when building releases that include
the Erlang client.

Fixes https://github.com/rabbitmq/rabbitmq-erlang-client/issues/72.

### DIFF
--- a/src/rabbit_common.app.src
+++ b/src/rabbit_common.app.src
@@ -9,6 +9,8 @@
 	{applications, [
 		kernel,
                 stdlib,
-                xmerl
+                xmerl,
+                syntax_tools,
+                compiler
 	]}
 ]}.


michaelklishin:
FTR, we had to do https://github.com/rabbitmq/rabbitmq-erlang-client/commit/712f7b341b3aac20f8fdbb23fd960aea4d0f1f0f as well.

essen:
They're application libraries, they do not start any process and therefore do not need to be started. At best they may need to be loaded but only if you set environment values (doesn't sound likely for these). But it doesn't hurt to have this (it'll be a load + noop).

michaelklishin:
@essen yes but we have a piece of code in the Erlang client that explicitly checks that this app (`rabbbit_common`) is running and without the above commit we observed failures in `amqp_client` startup on CI.

essen:
Alright, missed that, thanks.

PR #151, opened 2016-12-01
Configurable shutdown timeouts

hairyhum:
Fixes https://github.com/rabbitmq/rabbitmq-server/issues/847

Loads shutdown timeout for supervisors and workers from application environment.
Can have impact on performance. Therefore needs to be perf-tested.

### DIFF
--- a/include/rabbit.hrl
+++ b/include/rabbit.hrl
@@ -147,8 +147,10 @@
 -define(EMPTY_FRAME_SIZE, 8).
 
 -define(MAX_WAIT, 16#ffffffff).
--define(SUPERVISOR_WAIT, infinity).
--define(WORKER_WAIT, 30000).
+-define(SUPERVISOR_WAIT,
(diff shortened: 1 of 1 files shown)


hairyhum:
There is no significant performance impact of this change.

Tested with https://github.com/rabbitmq/rabbit-stress (creating lots of connections/channels/queues)

include/rabbit.hrl
```diff
@@ -147,8 +147,10 @@
 -define(EMPTY_FRAME_SIZE, 8).
 
 -define(MAX_WAIT, 16#ffffffff).
--define(SUPERVISOR_WAIT, infinity).
--define(WORKER_WAIT, 30000).
+-define(SUPERVISOR_WAIT,
+        application:get_env(rabbit, supervisor_shutdown_timeout, infinity)).
```
michaelklishin: `application:get_env/3` is not available in OTP R16B03. This should use `rabbit_misc:get_env/3`.

PR #154, opened 2016-12-08
rabbit_core_metrics: Restore compatibility with Erlang R16B03/17.x

dumbbell:
`ets:update_counter/4` was added to Erlang 18.x. Thus we need a wrapper to support Erlang down to R16B03.

This wrapper uses `ets:insert_new/2` first, then `ets:update_counter/3`. It lacks the atomicity of `ets:update_counter/4`, so if the entry was removed in between, the exception is simply ignored.

To avoid any penalty on Erlang 18+, we use `code_version` to load the appropriate version of the wrapper at runtime. Thus we still use `ets:udpate_counter/4` on Erlang 18+.

[#135769137]

### DIFF
--- a/src/rabbit_core_metrics.erl
+++ b/src/rabbit_core_metrics.erl
@@ -45,6 +45,22 @@
 
 -export([node_node_stats/2]).
 
+%% Those functions are exported for internal use only, not for public
+%% consumption.
+-export([
+         ets_update_counter/4,
+         ets_update_counter_pre_18/4,
+         ets_update_counter_post_18/4
+        ]).
+
+-erlang_version_support([
+                         {18, [
+                               {ets_update_counter, 4,
+                                ets_update_counter_pre_18,
+                                ets_update_counter_post_18}
+                              ]}
+                        ]).
+
 %%----------------------------------------------------------------------------
 %% Types
 %%----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


dumbbell:
This change is for `stable` only, it must not be merged to `master`!

src/rabbit_core_metrics.erl
```diff
@@ -130,35 +146,59 @@ channel_stats(reductions, Id, Value) ->
...
+%% ets:update_counter(Tab, Key, Incr, Default) appeared in Erlang 18.x.
+%% We need a wrapper for Erlang R16B03 and Erlang 17.x. This wrapper
+%% first creates the entry if it's missing, then updates it.
+
+ets_update_counter(Tab, Key, Incr, Default) ->
+    code_version:update(?MODULE),
+    ?MODULE:ets_update_counter(Tab, Key, Incr, Default).
+
+ets_update_counter_pre_18(Tab, Key, Incr, Default) ->
+    ets:insert_new(Tab, Default),
```
dcorbacho: I'd insert only if the `update_counter` fails, in the `badarg`, thus avoiding two ETS operations in most cases

dcorbacho: Not strictly.`ets:update_counter/4` always inserts the value if they key is not there, thus `update_counter/4` always succeeds. Now, we are allowing `0`s (let's do nothing) which weren't possible before.

dumbbell: You're right. I changed the patch to use your approach. Thanks!

PR #155, opened 2016-12-21
Topic-based authorisation for publishes

acogoluegnes:
Part of rabbitmq/rabbitmq-server#505.

### DIFF
--- a/include/rabbit.hrl
+++ b/include/rabbit.hrl
@@ -38,8 +38,8 @@
 -record(permission, {configure, write, read}).
 -record(user_vhost, {username, virtual_host}).
 -record(user_permission, {user_vhost, permission}).
--record(topic_key, {user_vhost, name}).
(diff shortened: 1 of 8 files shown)


include/rabbit.hrl
```diff
@@ -38,6 +38,8 @@
 -record(permission, {configure, write, read}).
 -record(user_vhost, {username, virtual_host}).
 -record(user_permission, {user_vhost, permission}).
+-record(topic_permission_key, {user_vhost, name}).
```
michaelklishin: I think this should be `exchange` instead. "Topic name" as a concept is a bit ambiguous in the current codebase (and AMQP 0-9-1). So until we get to the promised "protocol agnostic core" land it'd probably be clearer as `exchange`, in particular since the docs use the same name.

Note that CLI already uses "exchange" everywhere.

PR #157, opened 2017-01-05
Introduce credential validation

michaelklishin:
Part of https://github.com/rabbitmq/rabbitmq-server/issues/1054.


src/rabbit_auth_backend_internal.erl
```diff
@@ -165,7 +165,19 @@ permission_index(read)      -> #permission.read.
 %%----------------------------------------------------------------------------
 %% Manipulation of the user database
 
+validate_credentials(_Username, Password) ->
+    rabbit_credential_validation:validate_password(Password).
```
michaelklishin: If we pass both then we need to also rename the built-in validators to reflect that they only validate passwords. OK.

PR #160, opened 2017-01-19
Add topic authorisation for consumption

acogoluegnes:
Part of rabbitmq/rabbitmq-server#1085


michaelklishin:
This does not actually implement routing key checks on `queue.bind` and `queue.unbind` operations. I removed those references from the docs, we can do it later if it's considered to be essential. We have at least a couple of related plugins to update before `3.7.0 RC1` anyway and this can use some dogfooding in the meantime.

PR #161, opened 2017-01-23
Build *.ez archives for Elixir dependencies

hairyhum:
Detect dependencies with `mix.exs` file and build `*.ez` archives for them using the [Archive.Build.All](https://github.com/hairyhum/mix_task_archive_deps) Mix task.

The task lib is installed using `Mix.Archive.Install` if not present in `MIX_ARCHIVES` already.

Required for rabbitmq/rabbitmq-auth-backend-uaa#5

rabbitmq/rabbitmq-auth-backend-uaa#3

### DIFF
--- a/mk/rabbitmq-dist.mk
+++ b/mk/rabbitmq-dist.mk
@@ -1,10 +1,16 @@
 .PHONY: dist test-dist do-dist clean-dist
 
 DIST_DIR = plugins
+MIX = echo y | mix
 
 dist_verbose_0 = @echo " DIST  " $@;
 dist_verbose_2 = set -x;
 dist_verbose = $(dist_verbose_$(V))
+
+MIX_ARCHIVES ?= $(HOME)/.mix/archives
+
+MIX_TASK_ARCHIVE_DEPS_VERSION = 0.2.0
+mix_task_archive_deps = $(MIX_ARCHIVES)/mix_task_archive_deps-$(MIX_TASK_ARCHIVE_DEPS_VERSION)
 
 # We take the version of an Erlang application from the .app file. This
 # macro is called like this:
@@ -22,15 +28,19 @@ $(shell awk '
 }' $(1))
 endef
 
-# Define the target to create an .ez plugin archive. This macro is
-# called like this:
+define get_mix_project_version
+$(shell cd $(1) && $(MIX) run -e "IO.puts(Mix.Project.config[:version])")
+endef
+
+# Define the target to create an .ez plugin archive for an
+# Erlang.mk-based project. This macro is called like this:
 #
-#   $(call do_ez_target,app_name,app_version,app_dir)
+#   $(call do_ez_target_erlangmk,app_name,app_version,app_dir)
 
-define do_ez_target
-dist_$(1)_ez_dir = $$(if $(2),$(DIST_DIR)/$(1)-$(2),$$(if $$(VERSION),$(DIST_DIR)/$(1)-$$(VERSION),$(DIST_DIR)/$(1)))
+define do_ez_target_erlangmk
+dist_$(1)_ez_dir = $$(if $(2),$(DIST_DIR)/$(1)-$(2), \
+	$$(if $$(VERSION),$(DIST_DIR)/$(1)-$$(VERSION),$(DIST_DIR)/$(1)))
 dist_$(1)_ez = $$(dist_$(1)_ez_dir).ez
-
 
 $$(dist_$(1)_ez): APP     = $(1)
 $$(dist_$(1)_ez): VSN     = $(2)
@@ -52,14 +62,42 @@ $$(dist_$(1)_ez): $$(patsubst %,$(3)/%, \
 	APP=$(1) VSN=$(2) EZ_DIR=$$(abspath $$(dist_$(1)_ez_dir))))
 endif
 
-DIST_EZS += $$(dist_$(1)_ez)
+ERLANGMK_DIST_EZS += $$(dist_$(1)_ez)
+
+endef
+
+# Define the target to create an .ez plugin archive for a Mix-based
+# project. This macro is called like this:
+#
+#   $(call do_ez_target_mix,app_name,app_version,app_dir)
+
+define get_mix_project_dep_ezs
+$(shell cd $(1) && \
+	$(MIX) do deps.get, deps.compile > /dev/null && \
+	$(MIX) archive.build.all.list -e -o $(DIST_DIR))
+endef
+
+define do_ez_target_mix
+dist_$(1)_ez_dir = $$(if $(2),$(DIST_DIR)/$(1)-$(2), \
+	$$(if $$(VERSION),$(DIST_DIR)/$(1)-$$(VERSION),$(DIST_DIR)/$(1)))
+dist_$(1)_ez = $$(dist_$(1)_ez_dir).ez
+
+$$(dist_$(1)_ez): APP     = $(1)
+$$(dist_$(1)_ez): VSN     = $(2)
+$$(dist_$(1)_ez): SRC_DIR = $(3)
+$$(dist_$(1)_ez): EZ_DIR  = $$(abspath $$(dist_$(1)_ez_dir))
+$$(dist_$(1)_ez): EZ      = $$(dist_$(1)_ez)
+
+MIX_DIST_EZS += $$(dist_$(1)_ez)
+EXTRA_DIST_EZS += $$(call get_mix_project_dep_ezs,$(3))
 
 endef
 
 # Real entry point: it tests the existence of an .app file to determine
 # if it is an Erlang application (and therefore if it should be provided
-# as an .ez plugin archive). Then, if calls do_ez_target. It should be
-# called as:
+# as an .ez plugin archive) and calls do_ez_target_erlangmk. If instead
+# it finds a Mix configuration file, it calls do_ez_target_mix. It
+# should be called as:
 #
 #   $(call ez_target,app_name)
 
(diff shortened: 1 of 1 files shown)


mk/rabbitmq-dist.mk
```diff
@@ -123,23 +163,34 @@ $(DIST_DIR)/%.ez:
...
 		$(MAKE) --no-print-directory -C $(SRC_DIR) prepare-dist \
 		APP=$(APP) VSN=$(VSN) EZ_DIR=$(EZ_DIR)
-	$(verbose) (cd $(DIST_DIR) && $(ZIP) $(ZIP_V) -r $*.ez $*)
+	$(verbose) (cd $(DIST_DIR) && $(ZIP) $(ZIP_V) -r `basename $(EZ)` `basename $(EZ) .ez`)
 	$(verbose) rm -rf $(EZ_DIR)
 
+$(MIX_DIST_EZS): $(mix_task_archive_deps_installed)
+	$(verbose) rm -rf $(DIST_DIR)/tmp
+	$(verbose) mkdir $(DIST_DIR)/tmp
+	$(verbose) cd $(SRC_DIR) && MIX do deps.get, deps.compile, compile, archive.build.all -e -o $(realpath $(DIST_DIR))/tmp
```
dumbbell: Could you please split that long line?

mk/rabbitmq-dist.mk
```diff
@@ -123,23 +163,34 @@ $(DIST_DIR)/%.ez:
...
+
+$(mix_task_archive_deps_installed):
+	mix archive.install --force https://github.com/hairyhum/mix_task_archive_deps/releases/download/0.1.0/mix_task_archive_deps-0.1.0.ez
```
dumbbell: Likewise, could you please split that line?

mk/rabbitmq-dist.mk
dumbbell: You can use `$(dist_verbose)` for this line so that the build is reported in the default case.

mk/rabbitmq-dist.mk
dumbbell: Could you please prepend `$(verbose)`?

mk/rabbitmq-dist.mk
```diff
@@ -123,23 +163,34 @@ $(DIST_DIR)/%.ez:
...
+
 # We need to recurse because the top-level make instance is evaluated
 # before dependencies are downloaded.
 
 MAYBE_APPS_LIST = $(if $(shell test -f $(ERLANG_MK_TMP)/apps.log && echo OK),$(ERLANG_MK_TMP)/apps.log)
 
 dist:: $(ERLANG_MK_RECURSIVE_DEPS_LIST) all
+	$(gen_verbose) rm -rf $(DIST_DIR)/*.ez
```
dumbbell: There can only be a single `$(*_verbose)` line, otherwise the target will be reported as if it was built twice. You can change the next line to `($verbose)`.

dumbbell: If only files are expected here, could you please remove the `-r` flag?

mk/rabbitmq-dist.mk
```diff
@@ -123,23 +163,34 @@ $(DIST_DIR)/%.ez:
...
 	$(gen_verbose) $(MAKE) do-dist DIST_PLUGINS_LIST="$(ERLANG_MK_RECURSIVE_DEPS_LIST) $(MAYBE_APPS_LIST)"
 
 test-dist:: $(ERLANG_MK_RECURSIVE_TEST_DEPS_LIST) test-build
 	$(gen_verbose) $(MAKE) do-dist DIST_PLUGINS_LIST="$(ERLANG_MK_RECURSIVE_TEST_DEPS_LIST) $(MAYBE_APPS_LIST)"
 
 do-dist:: $(DIST_EZS)
-	$(verbose) unwanted='$(filter-out $(DIST_EZS),$(wildcard $(DIST_DIR)/*.ez))'; \
+	$(verbose) unwanted='$(filter-out $(foreach archive,$(wildcard $(DIST_DIR)/*.ez),$(DIST_DIR)/$(notdir $(archive))),$(filter-out $(DIST_EZS),$(wildcard $(DIST_DIR)/*.ez)))'; \
```
dumbbell: I don't understand that line, can you explain it?

mk/rabbitmq-dist.mk
```diff
@@ -2,10 +2,22 @@
...
+
 dist_verbose_0 = @echo " DIST  " $@;
 dist_verbose_2 = set -x;
 dist_verbose = $(dist_verbose_$(V))
 
+ifeq ($(MIX_ARCHIVES),)
+MIX_ARCHIVES = $(wildcard ~/.mix/archives)
+endif
+
+mix_task_archive_deps_installed = $(lastword $(wildcard $(MIX_ARCHIVES)/mix_task_archive_deps-*))
```
hairyhum: I didn't tested it.

mk/rabbitmq-dist.mk
```diff
@@ -123,23 +163,34 @@ $(DIST_DIR)/%.ez:
 		&& grep -q '^prepare-dist::' $(SRC_DIR)/Makefile) || \
 		$(MAKE) --no-print-directory -C $(SRC_DIR) prepare-dist \
 		APP=$(APP) VSN=$(VSN) EZ_DIR=$(EZ_DIR)
-	$(verbose) (cd $(DIST_DIR) && $(ZIP) $(ZIP_V) -r $*.ez $*)
+	$(verbose) (cd $(DIST_DIR) && $(ZIP) $(ZIP_V) -r `basename $(EZ)` `basename $(EZ) .ez`)
```
hairyhum: We don't have a matched pattern here (`$(ERLANG_DIST_EZS)` vs `$(DIST_DIR)/%.ez`)

mk/rabbitmq-dist.mk
hairyhum: It makes no sense. There was this "delete all" line in the `dest` target: `$(gen_verbose) rm -rf $(DIST_DIR)/*.ez`, which we probably should keep.
The problem is that MIX dependencies build their dependencies by themselves and it's not easy to get the unwanted list.

hairyhum:
To clean only unwanted archives and leave wanted we can extend the archive.build mix task, to print the dependencies archives

mk/rabbitmq-dist.mk
```diff
@@ -70,9 +106,12 @@ dist_$(1)_appdir = $$(if $$(filter $(PROJECT),$(1)), \
 			      $(APPS_DIR)/$(1), \
 			      $(DEPS_DIR)/$(1)))
 dist_$(1)_appfile = $$(dist_$(1)_appdir)/ebin/$(1).app
+dist_$(1)_mixfile = $$(dist_$(1)_appdir)/mix.exs
 
 $$(if $$(shell test -f $$(dist_$(1)_appfile) && echo OK), \
-  $$(eval $$(call do_ez_target,$(1),$$(call get_app_version,$$(dist_$(1)_appfile)),$$(dist_$(1)_appdir))))
+  $$(eval $$(call do_ez_target,$(1),$$(call get_app_version,$$(dist_$(1)_appfile)),$$(dist_$(1)_appdir))), \
+  $$(if $$(shell test -f $$(dist_$(1)_mixfile) && [ "x$(1)" != "xrabbitmqctl" ] && [ "x$(1)" != "xrabbitmq_cli" ] && echo OK), \
```
hairyhum: I hope we can. It is scary.
I've tried to use conditions in functions, but it didn't worked.

mk/rabbitmq-dist.mk
dumbbell: Ok, I will test.

Just to be more specific, I had issues with `$(wildcard)` when dealing with an absolute path in the home directory. There is no issue with other `$(wildcard)` limited to a relative subdirectory of the project because we only use US-ASCII there.

mk/rabbitmq-dist.mk
```diff
@@ -123,23 +162,32 @@ $(DIST_DIR)/%.ez:
 		&& grep -q '^prepare-dist::' $(SRC_DIR)/Makefile) || \
 		$(MAKE) --no-print-directory -C $(SRC_DIR) prepare-dist \
 		APP=$(APP) VSN=$(VSN) EZ_DIR=$(EZ_DIR)
-	$(verbose) (cd $(DIST_DIR) && $(ZIP) $(ZIP_V) -r $*.ez $*)
+	$(verbose) (cd $(DIST_DIR) && $(ZIP) $(ZIP_V) -r $(notdir $(EZ)) $(notdir $(EZ_DIR)))
```
dumbbell: What was wrong with `$*`?

mk/rabbitmq-dist.mk
```diff
@@ -123,23 +162,32 @@ $(DIST_DIR)/%.ez:
...
-	$(gen_verbose) $(MAKE) do-dist DIST_PLUGINS_LIST="$(ERLANG_MK_RECURSIVE_DEPS_LIST) $(MAYBE_APPS_LIST)"
+	$(gen_verbose) rm -f $(DIST_DIR)/*.ez
+	$(verbose) $(MAKE) do-dist DIST_PLUGINS_LIST="$(ERLANG_MK_RECURSIVE_DEPS_LIST) $(MAYBE_APPS_LIST)"
 
 test-dist:: $(ERLANG_MK_RECURSIVE_TEST_DEPS_LIST) test-build
 	$(gen_verbose) $(MAKE) do-dist DIST_PLUGINS_LIST="$(ERLANG_MK_RECURSIVE_TEST_DEPS_LIST) $(MAYBE_APPS_LIST)"
 
 do-dist:: $(DIST_EZS)
-	$(verbose) unwanted='$(filter-out $(DIST_EZS),$(wildcard $(DIST_DIR)/*.ez))'; \
-	test -z "$$unwanted" || (echo " RM     $$unwanted" && rm -f $$unwanted)
```
dumbbell: Could you please revert that change to keep the filtered rm? This helps a lot when debugging circular dependencies because it shows when a dependency we just built is removed a few commands later.

hairyhum: The problem is that filtered `rm` will delete elixir dependencies (of dependencies), because they are not listed by recursive dependencies search.

mk/rabbitmq-dist.mk
hairyhum: I've added `.SUFFIXES: .ez`, but now it matches `plugins/blah.ez`, not just `blah.ez`, so we'll  have to use `$(notdir ..)` anyway.

mk/rabbitmq-dist.mk
hairyhum: `MIX_ARCHIVES` is located in home directory by default. But we can change that.

mk/rabbitmq-dist.mk
dumbbell: Where did you add `.SUFFIXES`?

dumbbell: Sorry, I understand the issue now, after looking at the whole file instead of just this diff chunk.

mk/rabbitmq-dist.mk
hairyhum: Now Mix can list the dependencies and we can skip them when removing unwanted.

hairyhum:
I hope that's it with correctness. It still could have some refactoring.

mk/rabbitmq-dist.mk
dumbbell: Thanks for adding the new mix target. I came to about the same patch to `rabbitmq-dist.mk` locally. I also update various comments which were no longer matching the actual code.

The only non-trivial difference we have is the call to `$(MIX) do deps.get, deps.compile, compile` before calling `archive.build.all.list`. I understand that `deps.get` is required to fetch dependencies and resolve their final version. However, do we need to compile them? Also, can we do `deps.get` and `archive.build.all.list` in the same mix command?

hairyhum: `deps.compile` is required to get versions for deps. Mix doesn't report a version for a non-compiled dependency. 
We probably don't need `compile` task for listing dependencies.
We cannot do `deps.get` and `archive.build.all.list` in the same command, because we don't want output from `deps.get`

dumbbell:
I pushed some changes, mainly in comments. I tested it successfully on Windows too.

Issue #162, opened 2017-01-24
Precondition_failed formatting error with unicode queue name


Gsantomaggio:
`precondition_failed` raises an `badarg` error when the queue name is in unicode:

```
Error: badarg
Stack trace: [{io_lib,format,
                  ["Channel error on connection ~p (~s, vhost: '~s', user: '~s'), channel ~p:~n~s~n",
                   [<0.507.0>,<<"127.0.0.1:50536 -> 127.0.0.1:5672">>,<<"/">>,
                    <<"test">>,1,
                    [111,112,101,114,97,116,105,111,110,32,"queue.declare",32,
                     99,97,117,115,101,100,32,97,32,99,104,97,110,110,101,108,
                     32,101,120,99,101,112,116,105,111,110,32,
                     "precondition_failed",58,32,
                     [91,
                      ["105",44,10,
```

[this](https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_channel.erl#L682)  should be:
`io_lib:format("operation ~s caused a channel exception ~s: ~ts", [M, N, E]);`

as result:

```
=ERROR REPORT==== 24-Jan-2017::11:14:56 ===
Channel error on connection <0.462.0> (127.0.0.1:50816 -> 127.0.0.1:5672, vhost: '/', user: 'test'), channel 1:
operation queue.declare caused a channel exception precondition_failed: invalid arg 'x-queue-mode' 
for queue '卓匃屟尢桨' in vhost '/': invalid_queue_mode
```

PR #163, opened 2017-01-24
Change the message format

Gsantomaggio:
Fixes:https://github.com/rabbitmq/rabbitmq-common/issues/162

to reproduce the problem:

```
curl -i -u guest:guest -H "content-type:application/json" \
  -XPUT -d'{"auto_delete":false,"durable":false,"arguments":{"x-queue-mode":"wrong"}}' \
  http://localhost:15672/api/queues/%2f/卓匃屟尢桨
```

### DIFF
--- a/src/rabbit_channel.erl
+++ b/src/rabbit_channel.erl
@@ -660,7 +660,7 @@ send(Command, #ch{writer_pid = WriterPid}) ->
     ok = rabbit_writer:send_command(WriterPid, Command).
 
 format_soft_error(#amqp_error{name = N, explanation = E, method = M}) ->
-    io_lib:format("operation ~s caused a channel exception ~s: ~p", [M, N, E]);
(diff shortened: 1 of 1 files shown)


src/rabbit_channel.erl
```diff
@@ -660,7 +660,7 @@ send(Command, #ch{writer_pid = WriterPid}) ->
     ok = rabbit_writer:send_command(WriterPid, Command).
 
 format_soft_error(#amqp_error{name = N, explanation = E, method = M}) ->
-    io_lib:format("operation ~s caused a channel exception ~s: ~p", [M, N, E]);
+    io_lib:format("operation ~s caused a channel exception ~s: ~ts", [M, N, E]);
```
michaelklishin: For those wondering what the `t` is for: it's a Unicode translation modifier.

Issue #164, opened 2017-01-25
Add the Options parameter to rabbit_json 


Gsantomaggio:
The current implementation exposes only:

```
decode(JSON) ->
    jsx:decode(JSON, [return_maps]).
```

The JSX library supports different parameters for [`decode`](https://github.com/talentdeficit/jsx#decode12) and [ `encode`](https://github.com/talentdeficit/jsx#encode12) functions

It is useful to expose all these interfaces for future implementations and also for plugins that use the `rabbit_json`.

I am working on this [issue](https://github.com/aweber/rabbitmq-autocluster/issues/120) on  rabbitmq-autocluster, and I had to add use `jsx:decode` directly. See this [comment](https://github.com/aweber/rabbitmq-autocluster/issues/120#issuecomment-274853144) .

This is what I have in mind to do:
```
-spec decode(jsx:json_text()) -> jsx:json_term().
decode(JSON) ->
    decode(JSON, []).


-spec decode(jsx:json_text(), jsx_to_term:config()) -> jsx:json_term().
decode(JSON, Opts) ->
    jsx:decode(JSON, Opts).
```

**There is a little side effect:**

Currently, the other RabbitMQ modules use this interface:
```
rabbit_json:try_decode(Defn) %% this interface implicitly uses  `[return_maps]`
```

and it should be replaced with:
```
rabbit_json:try_decode(Defn, [return_maps]) 
```

I already tried to modify the interface by adding `[return_maps]`, then ran the unit tests on the management ui and they passed

michaelklishin:
@Gsantomaggio there's still time to update all callers for `3.7.0` :)

PR #165, opened 2017-01-26
Add the Options parameter

Gsantomaggio:
Part of https://github.com/rabbitmq/rabbitmq-common/issues/164.

### DIFF
--- a/src/rabbit_json.erl
+++ b/src/rabbit_json.erl
@@ -16,20 +16,37 @@
 
 -module(rabbit_json).
 
--export([decode/1, try_decode/1, encode/1]).
+-export([decode/1, decode/2, try_decode/1, try_decode/2, encode/1, encode/2]).
+
 
 -spec decode(jsx:json_text()) -> jsx:json_term().
 decode(JSON) ->
-    jsx:decode(JSON, [return_maps]).
+    decode(JSON, []).
+
+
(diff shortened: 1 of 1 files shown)


src/rabbit_json.erl
```diff
@@ -16,20 +16,53 @@
...
+
+-spec try_decode(jsx:json_text(), jsx_to_term:config()) -> 
+			{ok, jsx:json_term()} | error.
+try_decode(JSON, Opts) ->
     try
-        {ok, decode(JSON)}
+        {ok, decode(JSON, Opts)}
     catch error:_ ->
-        error
+	    error
```
Gsantomaggio: I was wondering if it makes sense to modify the error handling in this way:

```
  catch error: Reason ->
        {error, Reason}
    end.
```
So, in errors like [this](https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-common-164/src/rabbit_vhost_limit.erl#L140) we can add more info to understand the problem.

michaelklishin:
It does make sense.

michaelklishin:
So this is not 100% what I had in mind. `decode/1` and `encode/1` can stay opinionated (use `return_maps` or any other options we prefer), we should have only introduced new arities for plugins that really need them, such as `rabbitmq-autocluster`.

PR #168, opened 2017-01-31
Clean channel metrics when channel exits with an uncontrolled termination

### LINKED ISSUE #167: Channels aren't cleaned properly when connections are brutally killed
When rabbitmq-perf-test producers crashed with out-of-memory errors, 97 channels were left behind:

```
Error on AMQP connection <0.3361.0> (10.0.16.31:50086 -> 10.0.16.23:5672, vhost: '8961921b-36f6-48e0-9553-0351c7783376', user: '600878f0-8001-41e3-8198-986d941db2eb', state: running), channel 5:error while terminating:
killed
```

There were 97 instances of the above error in the rabbit log.

gerhard:
Fixes #167

### DIFF
--- a/src/rabbit_reader.erl
+++ b/src/rabbit_reader.erl
@@ -723,10 +723,12 @@ handle_dependent_exit(ChPid, Reason, State) ->
     {Channel, State1} = channel_cleanup(ChPid, State),
     case {Channel, termination_kind(Reason)} of
         {undefined,   controlled} -> State1;
-        {undefined, uncontrolled} -> exit({abnormal_dependent_exit,
+        {undefined, uncontrolled} -> handle_uncontrolled_channel_close(ChPid),
+                                     exit({abnormal_dependent_exit,
                                            ChPid, Reason});
         {_,           controlled} -> maybe_close(control_throttle(State1));
-        {_,         uncontrolled} -> State2 = handle_exception(
+        {_,         uncontrolled} -> handle_uncontrolled_channel_close(ChPid),
+                                     State2 = handle_exception(
                                                 State1, Channel, Reason),
                                      maybe_close(control_throttle(State2))
     end.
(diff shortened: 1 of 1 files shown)


src/rabbit_reader.erl
```diff
@@ -767,6 +767,7 @@ wait_for_channel_termination(N, TimerRef,
                                "error while terminating:~n~p~n",
                         [self(), ConnName, VHost, User#user.username,
                          CS, Channel, Reason]),
+                    rabbit_core_metrics:channel_closed(ChPid),
```
michaelklishin: Should a `rabbit_event` event be emitted here, since `terminate/2` for channel processes can't be relied on in the case of `brutal_kill` (or so it seems)? Like here: https://github.com/rabbitmq/rabbitmq-common/blob/b83f586974292cecf884c0369bbdcd3e2ed5edee/src/rabbit_channel.erl#L647.

dcorbacho: @michaelklishin you're right, to delete stats we listen to events. The core metrics only is not enough

michaelklishin: A test in `rabbitmq-management` would be great if we can find a way to make a channel terminate this way.

I suggest we finish this, merge it, and deploy into a long-running environment until we can figure out a way to do this.

kjnilsson: We've added the event notification as well.

src/rabbit_reader.erl
```diff
@@ -767,6 +767,8 @@ wait_for_channel_termination(N, TimerRef,
...
+                    rabbit_event:notify(channel_closed, [{pid, self()}]),
```
michaelklishin: I believe this should be `ChPid`, not `self`?

kjnilsson: Indeed! I'll fix that.

PR #169, opened 2017-02-02
Support proxy protocol

acogoluegnes:
Part of rabbitmq/rabbitmq-server#589


michaelklishin:
Heroku has merged our PR so we don't really have any reasons to bundle any more.

PR #170, opened 2017-02-02
mk/rabbitmq-components.mk: Switch to Cowboy 1.1.0

dumbbell:
The new feature of Cowboy 1.1.0 of interest to RabbitMQ is the possibility to serve static files from Erlang application .ez archives. This will allow us to get rid of the plugins expansion during RabbitMQ
startup.

Cowboy 1.1.0 also comes with a stricter URL parser. In particular, `+` characters are not converted to spaces anymore! They are left alone. Users of the HTTP API must make sure that their requests are encoded using path segment encoding, not query string encoding (where `+` are converted to spaces).

References ninenines/cowboy#1070.
References ninenines/cowboy#1080.
[#136779967]

### DIFF
--- a/mk/rabbitmq-components.mk
+++ b/mk/rabbitmq-components.mk
@@ -100,7 +100,7 @@ dep_rabbitmq_public_umbrella          = git_rmq rabbitmq-public-umbrella $(curre
 # all projects use the same versions. It avoids conflicts and makes it
 # possible to work with rabbitmq-public-umbrella.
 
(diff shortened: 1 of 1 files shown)


dumbbell:
FTR, the following issues were opened to fix URL encoding in HTTP clients:
* ruby-amqp/rabbitmq_http_api_client#28
* michaelklishin/rabbit-hole#95
* rabbitmq/hop#97

All of them were fixed already.

PR #171, opened 2017-02-03
New API for check_X_access to return error string

hairyhum:
Related to https://github.com/rabbitmq/rabbitmq-server/pull/1096
Part of https://github.com/rabbitmq/rabbitmq-auth-backend-uaa/issues/3

### DIFF
--- a/src/rabbit_authz_backend.erl
+++ b/src/rabbit_authz_backend.erl
@@ -47,7 +47,7 @@
 %%     Something went wrong. Log and die.
 -callback check_vhost_access(rabbit_types:auth_user(),
                              rabbit_types:vhost(), rabbit_net:socket()) ->
-    boolean() | {'error', any()}.
+    boolean() | {'error', any()} | {'error_message', string()}.
 
 %% Given #auth_user, resource and permission, can a user access a resource?
 %%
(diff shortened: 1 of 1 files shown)


src/rabbit_authz_backend.erl
```diff
@@ -47,7 +47,7 @@
 %%     Something went wrong. Log and die.
 -callback check_vhost_access(rabbit_types:auth_user(),
                              rabbit_types:vhost(), rabbit_net:socket()) ->
-    boolean() | {'error', any()}.
+    boolean() | {'error', any()} | {'error_message', string()}.
```
michaelklishin: We should explain the difference with a comment.

src/rabbit_authz_backend.erl
```diff
@@ -45,9 +45,11 @@
 %% false
 %% {error, Error}
 %%     Something went wrong. Log and die.
+%% {error_message, Message}
+%%     Report Message to a client using protocol error and die.
 -callback check_vhost_access(rabbit_types:auth_user(),
                              rabbit_types:vhost(), rabbit_net:socket()) ->
-    boolean() | {'error', any()}.
+    boolean() | {'error', any()} | {'error_message', string()}.
```
kjnilsson: Would `protocol_error` instead of `error_message` be clearer name here?

kjnilsson: I see from the other PR that both `{error, _}` and '{error_message, _}` result in a protocol error.

PR #183, opened 2017-02-28
Better error handling for module attributes

binarin:
Stacktrace elements can also be a 4-element tuples like
`{module,function,[attributes],[]}`

But as we only care about an `undef` error fact, it makes no sense
to inspect stacktrace contents.

### DIFF
--- a/src/rabbit_misc.erl
+++ b/src/rabbit_misc.erl
@@ -799,15 +799,13 @@ gb_trees_foreach(Fun, Tree) ->
     gb_trees_fold(fun (Key, Val, Acc) -> Fun(Key, Val), Acc end, ok, Tree).
 
 module_attributes(Module) ->
-    case catch Module:module_info(attributes) of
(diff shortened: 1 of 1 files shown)


michaelklishin:
@binarin thanks. How can I reproduce such a failure, just specify a non-existent module in a plugin boot step?

binarin:
You need to have some non-existing module mentioned in one of the loaded applications `modules`  attribute.

I can reproduce it by editing `ebin/rabbit.app` in a `rabbitmq-server` checkout and then `make run-broker` . Erlang version is 18.3.4

michaelklishin:
This works as expected and I agree that displaying a stack trace probably doesn't make much sense in this case.

binarin:
The problem was that the broker wasn't starting at all, as the changed stacktrace format prevented the first clause of original case from matching; and thus the second one fired the exit signal.

PR #186, opened 2017-03-07
Log username and vhost on successful connection.open

michaelklishin:
Fixes rabbitmq/rabbitmq-server#1140.


gerhard:
Happy as is, 👍  to merge

PR #189, opened 2017-04-17
Allow building plugins written in Elixir

binarin:
One of the thing RabbitMQ expects from a plugin is a dependency on the
`rabbit` app. The problem is that `mix run` tries to start plugin app,
which leads to an unsuccessful attempt to start `rabbit` app.


michaelklishin:
Thank you!

michaelklishin:
Thank you!

PR #190, opened 2017-04-26
Regenerate Elixir-based .ez-s after changes

binarin:
After changing any .erl-file and running `make dist` .ez files are
regenerated. But this is not the case for Elixir plugins - so it's
necessary to clean `plugins/` to refresh.

The difference is that for erlang.mk projects there is an explicit
dependencies on `ebin`, `include` and `priv` folder content -
https://github.com/rabbitmq/rabbitmq-common/blob/adaa9ac03943055ad68017df3b4272e8bf93bc3a/mk/rabbitmq-dist.mk#L52

But there were no similar rule around
https://github.com/rabbitmq/rabbitmq-common/blob/adaa9ac03943055ad68017df3b4272e8bf93bc3a/mk/rabbitmq-dist.mk#L91

So I've tried to add some reasonable dependencies there. I'm not so sure about `_build/dev/...` choice, but at least it works.


michaelklishin:
Merging because both @dumbbell and @hairyhum seem to approve :)

Issue #198, opened 2017-05-18
Integrate Looking Glass


essen:
When we run `make run-broker` we want to conditionally enable the Looking Glass tracer with trace pattern of our choosing. For example, `make run-broker RABBITMQ_TRACER=rabbit_tracer:connections` would trace all the connection processes. Then once tracing is done, `make profile` will build callgrind.out files that can then be opened with qcachegrind.

michaelklishin:
Using environment variables to trace one or more modules (or functions) would be great. Would 
a comma-separated list be sufficient here?

essen:
You provide a comma separated list of M:F which contain specifications for what you want to trace. More info in the coming PRs.

essen:
We also need to make Looking Glass "proper" with a LICENSE, README and whatnot. Same for the lz4-erlang dependency. Is there any guideline as to what I should add?

michaelklishin:
@essen I can help with those.

PR #199, opened 2017-05-18
Add Looking Glass support

essen:
Part of https://github.com/rabbitmq/rabbitmq-common/issues/198

We need to exclude looking_glass and lz4 from plugins and use
a different loading method because it is not possible to load
NIFs from within archives.

An example usage follows:

    make run-broker RABBITMQ_TRACER="rabbit_looking_glass:connections"

In another shell:

    bin/runjava com.rabbitmq.perf.PerfTest

Then some time later stop the test and the node, then run:

    make profile
    qcachegrind


essen:
Fixed a schoolboy error and pushed again.

PR #201, opened 2017-05-28
[supervisor2] Mark child restarting when delaying restart.

aartamonau:
Otherwise the child might not be restarted on failure under certain
circumstances (see the code to reproduce the issue below).

Here's the sequence of events leading to the issue.

 1. A child crashes multiple times forcing the supervisor2 to
    delay-restart it.

 2. Upon scheduling the delayed restart, the supervisor calls
    state_del_child/2 which sets child's pid to undefined.

 3. After the delay, on restart attempt the child crashes before
    replying to the supervisor.

 4. The supervisor schedules another restart attempt via
    try_again_restart/3. Before that restart/3 would try to mark the
    child as being restarted by means of calling restarting/1 on
    child's pid. But since the pid is undefined, the function keeps it
    unchanged.

 5. When the supervisor handles the try_again_restart cast, it expects
    the process to be in restarting state. Since it's not the case, it
    just ignores the message.

 6. As a result, the child doesn't get restarted.

This patch addresses the issue by marking the child as being restarted
in step 2. The only visible behavior change for the user is that when
the child is scheduled to be restarted, supervisor2:which_children
would return restarting instead of undefined for the child. This seems
appropriate and nobody should have relied on the old behavior.

The code to reproduce the issue:

    -module(reproducer).

    -behavior(supervisor2).

    -export([start/0, init/1, child/0]).

    start_link() ->
        supervisor2:start_link({local, parent}, ?MODULE, []).

    init([]) ->
        {ok, {{one_for_one, 1, 10},
              [{child, {proc_lib, start_link, [?MODULE, child, []]},
                {permanent, 2}, brutal_kill, worker, []}]}}.

    child() ->
        io:format("starting child~n"),

        case application:get_env(kernel, fail_in_init) of
            undefined ->
                ok;
            _ ->
                io:format("crashing in init~n"),
                exit(die)
        end,

        proc_lib:init_ack({ok, self()}),
        exit(die).

    start() ->
        {ok, _Sup} = start_link(),
        timer:sleep(500),
        application:set_env(kernel, fail_in_init, true),
        timer:sleep(10000),
        io:format("done~n"),
        init:stop().

Output without the patch:

    $ erl -noinput -run reproducer
    starting child
    starting child
    starting child
    crashing in init
    done

Output with the patch:

    $ erl -noinput -run reproducer
    starting child
    starting child
    starting child
    crashing in init
    starting child
    crashing in init
    starting child
    crashing in init
    starting child
    crashing in init
    starting child
    crashing in init
    done

### DIFF
--- a/src/supervisor2.erl
+++ b/src/supervisor2.erl
@@ -878,7 +878,8 @@ do_restart_delay({RestartType, Delay}, Reason, Child, State) ->
             _TRef = erlang:send_after(trunc(Delay*1000), self(),
                                       {delayed_restart,
                                        {{RestartType, Delay}, Reason, Child}}),
-            {ok, state_del_child(Child, State)}
+            OldPid = Child#child.pid,
+            {ok, replace_child(Child#child{pid=restarting(OldPid)}, State)}
     end.
 
 restart(Child, State) ->


michaelklishin:
Thank you for the PR and providing the details.

aartamonau:
The following change in the test fixes the problem:

```
diff --git i/test/sup_delayed_restart_SUITE.erl w/test/sup_delayed_restart_SUITE.erl
index e495f57d0..0dffe3306 100644
--- i/test/sup_delayed_restart_SUITE.erl
+++ w/test/sup_delayed_restart_SUITE.erl
@@ -81,6 +81,7 @@ exit_child(SupPid) ->
 with_child_pid(SupPid, Fun) ->
     case supervisor2:which_children(SupPid) of
         [{_Id, undefined, worker, [?MODULE]}] -> ok;
+        [{_Id, restarting, worker, [?MODULE]}] -> ok;
         [{_Id,  ChildPid, worker, [?MODULE]}] -> Fun(ChildPid);
         []                                     -> ok
     end.
```

Basically, it has to do with the fact that which_children now reports the child as being restarted. I mentioned this in my original commit message. Do you want me to change my patch or do you agree with me that that's a reasonable behavior?

michaelklishin:
@aartamonau it makes sense to me that `which_children/1` now reports children as restarting. We need to update the tests, so please amend your PR ;)

michaelklishin:
Addressed in https://github.com/rabbitmq/rabbitmq-server/pull/1238.

michaelklishin:
Backported in #202, should ship in `3.6.11`.

michaelklishin:
@aartamonau thank you for digging this one out and providing an example app to reproduce. This may be the root cause of a couple of long standing issues we didn't manage to reproduce, e.g. https://github.com/rabbitmq/rabbitmq-federation/issues/55 (also relevant: https://github.com/rabbitmq/rabbitmq-federation/issues/43, https://github.com/rabbitmq/rabbitmq-federation/issues/46).

PR #212, opened 2017-07-08
Use sysctl to get the total memory on Darwin (OS X)

gerhard:
Prior to this change, on a host with 16GB of RAM, RabbitMQ was reporting
total memory as `13719 MiB (14385987584 bytes)`. This was obviously
incorrect, the host has 16384MiB of RAM.

Other notable utilities that use sysctl to report total memory on OS X:

  * Ohai (Chef): https://github.com/chef/ohai/blob/v13.2.0/lib/ohai/plugins/darwin/memory.rb#L25
  * Facter (Puppet): https://github.com/puppetlabs/facter/blob/3.7.0/lib/src/facts/osx/memory_resolver.cc#L17

Using `/usr/bin/env sysctl` since:

  * OS X: `/usr/bin/sysctl`
  * FreeBSD & OpenBSD: `/sbin/sysctl`

### DIFF
--- a/src/vm_memory_monitor.erl
+++ b/src/vm_memory_monitor.erl
@@ -411,14 +411,7 @@ cmd(Command) ->
 %% Original code was part of OTP and released under "Erlang Public License".
 
 get_total_memory({unix,darwin}) ->
-    File = cmd("/usr/bin/vm_stat"),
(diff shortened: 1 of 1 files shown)


lukebakken:
works-on-my-machine

PR #213, opened 2017-07-10
Default memory high watermark to 1.0 when using a higher value

gerhard:
Prior to this change, setting the vm_memory_high_watermark to a value
higher than 1 (inclusive), would default to 0.4. This is different from
the behaviour when setting an absolute value, namely defaulting to the
total memory available.

Extracted common memory-related macros & records so that they can be
shared by tests.

re rabbitmq/rabbitmq-server#1285

[finishes #148470947]


include/rabbit_memory.hrl
```diff
@@ -0,0 +1,34 @@
...
+-define(ONE_MiB, 1048576).
+
+%% For an unknown OS, we assume that we have 1GB of memory. It'll be
+%% wrong. Scale by vm_memory_high_watermark in configuration to get a
+%% sensible value.
+-define(MEMORY_SIZE_FOR_UNKNOWN_OS, 1073741824).
+-define(DEFAULT_VM_MEMORY_HIGH_WATERMARK, 0.4).
+-define(MAX_VM_MEMORY_HIGH_WATERMARK, 1.0).
+
+-record(state, {total_memory,
```
michaelklishin: I'm afraid `state` is overly generic for a record in a shared header. Please rename this to `vm_mem_monitor_state` or `vm_memory_state`?

michaelklishin: In fact, since this record is only used in one module, we can move it to `vm_memory_monitor`.

PR #214, opened 2017-07-18
Report VM memory as system RSS only if it's bigger than erlang total.

hairyhum:
Related to rabbitmq/rabbitmq-server#1295
RSS memory can be smaller than `erlang:memory(total)`,
in that case it's safer to return memory reported by erlang.
[Addresses rabbitmq/rabbitmq-server#1294]
[#149066379]


michaelklishin:
Closing per discussion with @hairyhum.

PR #218, opened 2017-08-07
Revert the change of log qualifiers.

hairyhum:
Log qualifiers should be compatible with pre-3.7 qualfiers.
`rabbit_log_...` should be used only as fake module names
for lager parse-transform.

[#149634975]

### DIFF
--- a/src/rabbit_log.erl
+++ b/src/rabbit_log.erl
@@ -30,7 +30,7 @@
 -include("rabbit_log.hrl").
 %%----------------------------------------------------------------------------
 
--type category() :: atom().
+-type category() :: connection | channel | mirroring | queue | federation | upgrade.
 
 -spec debug(string()) -> 'ok'.
 -spec debug(string(), [any()]) -> 'ok'.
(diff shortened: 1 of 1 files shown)


src/rabbit_log.erl
```diff
@@ -73,14 +73,13 @@ log(Category, Level, Fmt, Args) when is_list(Args) ->
...
-make_internal_sink_name(rabbit_log_upgrade) -> rabbit_log_upgrade_lager_event;
-make_internal_sink_name(Category) ->
-    lager_util:make_internal_sink_name(Category).
+make_internal_sink_name(connection) -> rabbit_log_connection_lager_event;
+make_internal_sink_name(channel)    -> rabbit_log_channel_lager_event;
+make_internal_sink_name(mirroring)  -> rabbit_log_mirroring_lager_event;
+make_internal_sink_name(queue)      -> rabbit_log_queue_lager_event;
+make_internal_sink_name(federation) -> rabbit_log_federation_lager_event;
+make_internal_sink_name(upgrade)    -> rabbit_log_upgrade_lager_event;
+make_internal_sink_name(Category)   -> error({unknown_category, Category}).
```
dumbbell: Why making this an error, instead of calling `lager_utils:make_internal_sink_name()`?

And currently, if the category is unknown, the log message is lost; here is what is logged in my case:
```
2017-08-08 09:17:28.174 [error] <0.72.0> FORMAT ERROR: {unknown_category,connection2} []
```

hairyhum: The error here is to force people to use the defined set of categories. So running a test will show that the code has an error. Otherwise we would have lost logs and that's not what people generally want. If someone wants to have a custom sink, should it go through `rabbit_log:log`? I think they better use lager parse-transforms instead.
This function here is left for backwards compatibility and ideally should be deleted, for my opinion.

dumbbell: Ok, I understand and agree with you.

In this case, could you please fix the warning then?
```
src/rabbit_log.erl:82: Warning: ambiguous call of overridden auto-imported BIF error/1
 - use erlang:error/1 or "-compile({no_auto_import,[error/1]})." to resolve name clash
```

PR #219, opened 2017-08-29
Fix dep fetching on detached HEAD

binarin:
When non-English locale is used, detached HEAD magic stops working


michaelklishin:
Thank you!

PR #237, opened 2017-10-26
Start `rabbit` application as transient.

hairyhum:
This is a backport of #194 https://github.com/rabbitmq/rabbitmq-server/issues/1216
[#152310581]

### DIFF
--- a/src/app_utils.erl
+++ b/src/app_utils.erl
@@ -15,16 +15,19 @@
 %%
 -module(app_utils).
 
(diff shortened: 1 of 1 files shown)


gerhard:
Without this, killing `rabbit_sup` app results in 500 errors on the Management UI. It also leaves the Erlang node running, heart-beating as if nothing happened. This definitely addresses the issues that we've been seeing in the wild.

PR #241, opened 2017-11-29
WIP: Add a macro to identify processes in Looking Glass

essen:


pivotal-issuemaster:
@essen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-common&pullRequestId=241)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-common&pullRequestId=241) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@essen Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-common&pullRequestId=241)!

src/rabbit_writer.erl
```diff
@@ -175,6 +175,7 @@ system_code_change(Misc, _Module, _OldVsn, _Extra) ->
     {ok, Misc}.
 
 enter_mainloop(Identity, State) ->
+    ?LG_INFO(#{process_type => writer}),
```
hairyhum: Would it be more convenient to have `process_type` embedded into a macro? Since it's the only possible field currently. Are there any other potential values, which will be added to this map?

michaelklishin: Anything we choose to collect that's RabbitMQ-specific, for example.

essen: This value is the only one Looking Glass itself knows about.

On the other hand you could write a RabbitMQ-specific profiler/analyzer that would need more information and that's where you'd put it, because even if that tool could get the same info from other events (like function calls) it's a lot more solid to do it that way rather than have the tool break because the source code changed.

I don't expect us to add too many values there but I expect a few eventually when LG gets used for more RabbitMQ-specific things.

hairyhum: If we plan to add more fields to this map - then fine. Otherwise I would prefer a macro like this
```erlang
-define(LG_PROCESS_TYPE(Type), is_pid(whereis(lg)) andalso (lg ! #{process_type => Type})).

?LG_PROCESS_TYPE(reader)
```

essen: Let's add both macros then.

essen:
Done and rebased against current master.

PR #244, opened 2017-12-07
Add `rabbitmq-mix.mk` which is copied to Elixir components when we create the source archive

dumbbell:
References rabbitmq/rabbitmq-server-release#61.


dumbbell:
The commit to `rabbitmq-components.mk` is unrelated to this pull request. I'm not sure why GitHub shows it. I remember I made a mistake with this change and had to cherry-pick the change from `master`. I created a branch from `v3.7.x` to cherry-pick and merged it to `master`, to show the intent. But apparently, GitHub doesn't pick it.

PR #247, opened 2018-01-22
Extend amqqueue record to include queue type

dcorbacho:
Part of the new high availability implementation using raft as a consensus algorithm

Type defaults to 'classic'

[#154472130]


michaelklishin:
This is a 4.0-specific change, which means we either have to roll it back or create a branch for 3.8 and reconfigure Concourse and so on to not assume that master targets 3.8.0.

PR #254, opened 2018-03-06
Hide common_test node (3.6.x)

dumbbell:
As commented in `rabbitmq-early-test.mk`:

> We start the common_test node as a hidden Erlang node. The benefit is that other Erlang nodes won't try to connect to each other after discovering the common_test node if they are not meant to.
>
> This helps when several unrelated RabbitMQ clusters are started in parallel.

### DIFF
--- a/mk/rabbitmq-early-test.mk
+++ b/mk/rabbitmq-early-test.mk
@@ -23,6 +23,15 @@ endif
 # Common Test flags.
 # --------------------------------------------------------------------
 
+# We start the common_test node as a hidden Erlang node. The benefit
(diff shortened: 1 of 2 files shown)


dumbbell:
I'm particularily interested in comments on my change to `rabbit_misc:is_process_alive()`. I believe it's harmless, but I could be missing something.

michaelklishin:
If the node is not currently connected and automatic connection is disabled, the behavior will now be different (an early return instead of a timeout). Should be no different for our needs in practice unless I'm also missing something.

PR #258, opened 2018-03-06
Delete metrics for all deleted queues in 4 ETS operations (v3.7.x)

gerhard:
Rather than using 6 ETS operations per deleted queue, build a match
spec for every 1000 queues that uses `ets:select_replace/2` to delete
metrics in fewer ETS operations.

Great suggestions @michaelklishin & @hairyhum!
https://github.com/rabbitmq/rabbitmq-server/commit/ee0951b1b3c202eb261ad1830f28ab81904a7334#comments

This should be merged together with rabbitmq/rabbitmq-server#1526

For initial context, see rabbitmq/rabbitmq-server#1513.

Partner-in-crime: @essen

### DIFF
--- a/src/rabbit_core_metrics.erl
+++ b/src/rabbit_core_metrics.erl
@@ -265,28 +265,34 @@ delete_queue_metrics(Queue) ->
     ok.
 
 delete_channel_queue_exchange_metrics(MatchSpecCondition) ->
-    ets:select_replace(
+    ChannelQueueExchangeMetricsToUpdate = ets:select(
         channel_queue_exchange_metrics,
         [
             {
-                {{'$2', {'$1', '$3'}}, '$4', '_'},
+                {{'$2', {'$1', '$3'}}, '_', '_'},
                 [MatchSpecCondition],
-                [{{{{'$2', {{'$1', '$3'}}}}, '$4', 1}}]
+                [{{'$2', {{'$1', '$3'}}}}]
             }
         ]
-    ).
+    ),
+    lists:foreach(fun(Key) ->
+        ets:update_element(channel_queue_exchange_metrics, Key, {3, 1})
+    end, ChannelQueueExchangeMetricsToUpdate).
 
 delete_channel_queue_metrics(MatchSpecCondition) ->
-    ets:select_replace(
+    ChannelQueueMetricsToUpdate = ets:select(
         channel_queue_metrics,
         [
             {
-                {{'$2', '$1'}, '$3', '$4', '$5', '$6', '$7', '$8', '_'},
+                {{'$2', '$1'}, '_', '_', '_', '_', '_', '_', '_'},
                 [MatchSpecCondition],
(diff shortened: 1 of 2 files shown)


essen:
Efficiency is only a secondary concern. The main reason why this isn't a tail recursive function is that it would be really awkward to write one for building this kind of data structure.

michaelklishin:
Can anyone think of a scenario where `build_match_spec_conditions_to_delete_all_queues/1` would need to recurse deeply? If not then it's perfectly acceptable.

essen:
It will recurse once per queue and we've tested it so far with 50000 queues removed at once I think.

Note that body-recursion is no longer considered a problem as long as it stops looping at some point. See http://erlang.org/doc/efficiency_guide/myths.html#id62685 for more details.

michaelklishin:
In other words, it loops once per 50K queues? That's 100 times for 5M, sounds acceptable to me.

essen:
It loops once per queue. So in the case I mention it looped 50000 times. But the link I gave is more useful, there's no real difference between body and tail recursion for these functions as far as memory goes.

michaelklishin:
Memory is less of an issue. Stack overflow is. It would be interesting to try this with 5M iterations (note: not necessarily 5M actual queues, we can export the function and run it from a REPL session).

essen:
I don't believe there are stack overflows in Erlang. The stack just consumes more memory and as long as Erlang can allocate memory then it will keep running. In a body-recursive function the extra memory is used by the stack; in a tail-recursive function the extra memory is used by the accumulator; and the total memory used is equivalent.

Looking around quickly I found this link that gives a more detailed explanation: https://stackoverflow.com/questions/27028824/erlang-stackoverflow-with-recursive-function-that-is-not-tail-call-optimized

michaelklishin:
OK, I see what you mean now. Let’s test the function in isolation with 5M iterations to make a more informed decision.

It would be interesting to see what kind of difference it makes for a shutdown of a node with 50K queues. I have no doubt this is an improvement, just curious about the extent :)

essen:
Ah right we didn't give the good news. This change along with the others makes the cleanup almost instantaneous.

There's still some slowness when you stop a node normally (because the node being stopped stops queue processes one at a time and they cleanup after themselves) but once the node is finally stopped (leaving 90% of queues to be cleaned up, because of shutdown timeouts) the other node picks up the work and finishes cleaning up in a couple seconds.

We still have some experiments to do including with 3 nodes in the cluster but the changes so far are really promising.

michaelklishin:
`ets:select_replace/2` is a NIF so we cannot just copy it from OTP 20. An alternative code path for 19.3 will be necessary.

gerhard:
No worries, we'll come up with a solution that works across all supported versions.

@essen 🛠

essen:
Yes...

We'll have to go back to our initial version without select_replace, I don't think the difference in performance was very big anyway.

michaelklishin:
This code isn't on the hot code path so we potentially can delegate to the `ets:select_replace/2` version if it's available. The overhead of one or two function calls should not really matter here.

Issue #27, opened 2015-11-20
Missing "sslsocket" causing build failure


palfrey:
Previously (as in within the last 2 weeks) I was able to have rabbitmq-common as a dependency of my plugin project and the build worked. Now, it doesn't appear to work, and certainly when checking out current master on it's own I get build failures. Entirely possible this is a bug in my local system, or that I'm missing a dependency, but I can't see anything obvious. Might be related to #8 given where the issues are occurring, but that's a complete guess.

Any thoughts?

```
root@ae0389c87456:/rabbit/rabbitmq-common# make
 DEP    rabbitmq_codegen
 GEN    include/rabbit_framing.hrl
 GEN    src/rabbit_framing_amqp_0_8.erl
 GEN    src/rabbit_framing_amqp_0_9_1.erl
 DEPEND rabbit_common.d
 ERLC   app_utils.erl credit_flow.erl gen_server2.erl mirrored_supervisor.erl mochijson2.erl pmon.erl priority_queue.erl rabbit_amqqueue.erl rabbit_auth_mechanism.erl rabbit_authn_backend.erl rabbit_authz_backend.erl rabbit_backing_queue.erl rabbit_basic.erl rabbit_binary_generator.erl rabbit_binary_parser.erl rabbit_channel.erl rabbit_channel_interceptor.erl rabbit_command_assembler.erl rabbit_control_misc.erl rabbit_event.erl rabbit_exchange_decorator.erl rabbit_exchange_type.erl rabbit_framing_amqp_0_8.erl rabbit_framing_amqp_0_9_1.erl rabbit_heartbeat.erl rabbit_misc.erl rabbit_msg_store_index.erl rabbit_net.erl rabbit_networking.erl rabbit_nodes.erl rabbit_password_hashing.erl rabbit_policy_validator.erl rabbit_queue_collector.erl rabbit_queue_decorator.erl rabbit_queue_master_locator.erl rabbit_reader.erl rabbit_runtime_parameter.erl rabbit_writer.erl ssl_compat.erl supervisor2.erl time_compat.erl
src/rabbit_net.erl:27: can't find include lib "ssl/src/ssl_api.hrl"
src/rabbit_net.erl:99: record sslsocket undefined
src/rabbit_net.erl:109: record sslsocket undefined
src/rabbit_net.erl:114: record sslsocket undefined
src/rabbit_net.erl:119: record sslsocket undefined
src/rabbit_net.erl:124: record sslsocket undefined
[...]

PR #276, opened 2018-10-09
Remove dependency on erlang-src

essen:
It has been reported that in order to use the Erlang client, the
Erlang/OTP source must be available. This is due to one include
file that rabbit_net required. This dependency has been removed.

Instead of calling is_record(sslsocket) the macro ?IS_SSL will
now perform the same test manually (check that it is a tuple,
that the size is correct and that the first element equals sslsocket).

The tuple has not changed in a very long time so doing this
manually is at least as safe as including this private header
file (it could be removed or moved at any time).

Once Erlang/OTP 22 gets out and we know how sockets will be
represented with the NIF implementation, we could revise this
and check whether the socket is one that gen_tcp accepts
(currently it's a port, but this will probably change when
a NIF is used).

Fixes https://github.com/rabbitmq/rabbitmq-common/issues/27 - this should remove a pain point for many people for using the Erlang client.

Tested against some of the rabbitmq-server test suites that deal with ssl (proxy_protocol is one).

### DIFF
--- a/src/rabbit_net.erl
+++ b/src/rabbit_net.erl
@@ -18,7 +18,6 @@
 -include("rabbit.hrl").
 
 -include_lib("kernel/include/inet.hrl").
--include_lib("ssl/src/ssl_api.hrl").
 
 -export([is_ssl/1, ssl_info/1, controlling_process/2, getstat/2,
          recv/1, sync_recv/2, async_recv/3, port_command/2, getopts/2,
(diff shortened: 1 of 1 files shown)


michaelklishin:
This makes the code less or non-forward compatible but it sounds like OTP 22 will break anyway.

essen:
Note that my upcoming changes to use the PROXY protocol implementation in Ranch might make it possible to switch that macro to a simple `is_tuple`, since it won't wrap the socket like `ranch_proxy_protocol` is doing.

dumbbell:
Yes, I'm in favor of this change in order to get rid of that dependency on a private header.

PR #280, opened 2018-10-24
Add member/2 to priority_queue module

acogoluegnes:
To check whether a value is in the queue or not. Handle both
non-priority and priority queue. Based on lists:member/2, does not take
into account the priority of the value in the queue to check equality.

References rabbitmq/rabbitmq-server#1743

### DIFF
--- a/src/priority_queue.erl
+++ b/src/priority_queue.erl
@@ -11,7 +11,7 @@
 %% The Original Code is RabbitMQ.
 %%
 %% The Initial Developer of the Original Code is GoPivotal, Inc.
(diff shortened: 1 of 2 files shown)


src/priority_queue.erl
```diff
@@ -214,10 +216,28 @@ highest({queue, [], [], 0})     -> empty;
 highest({queue, _, _, _})       -> 0;
 highest({pqueue, [{P, _} | _]}) -> maybe_negate_priority(P).
 
+member(_X, {queue, [], [], 0}) ->
```
michaelklishin: What's the purpose of this function? When is something a "member" of a priority queue?

michaelklishin:
Superseded by #288.

PR #281, opened 2018-11-12
rabbitmq-run.mk: Fix stop-brokers with NODES > 1

jeckersb:
Remove extra "1" in seq command.  Previously, as an example with
NODES=2, will run `seq 2 1` which produces no items to iterate, so the
entire stop-node loop does not execute and the brokers are left
running.

(This should qualify as an Obvious Fix for CLA purposes)

### DIFF
--- a/mk/rabbitmq-run.mk
+++ b/mk/rabbitmq-run.mk
@@ -341,7 +341,7 @@ start-brokers start-cluster:
 	done
 
 stop-brokers stop-cluster:
-	@for n in $$(seq $(NODES) 1); do \
+	@for n in $$(seq $(NODES) -1 1); do \
 		nodename="rabbit-$$n@$$(hostname -s)"; \
 		$(MAKE) stop-node \
 		  RABBITMQ_NODENAME="$$nodename"; \


pivotal-issuemaster:
@jeckersb Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-common&pullRequestId=281)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-common&pullRequestId=281) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jeckersb This Pull Request contains an obvious fix. Signing the Contributor License Agreement is not necessary.

jeckersb:
@dumbbell thanks, makes much more sense to stop them in reverse order :)

dumbbell:
Thank you, I merged your patch.

That was my mistake. I only tested this on my workstation running FreeBSD and seq(1) on BSD doesn't have the same default behavior regarding the implicit increment as the GNU implementation. So the usual Works On My Machine™ :-)

PR #294, opened 2019-01-16
Introduce a new type, heartbeat_timeout, for documentation purposes

michaelklishin:
Per discussion with @lukebakken.


michaelklishin:
Backported to `v3.7.x`.

PR #300, opened 2019-02-07
Handle empty binary payload when delivering message

### LINKED ISSUE #299: Broker crashes when delivering message with empty binary payload
See https://groups.google.com/d/msg/rabbitmq-users/txOWyh05GqQ/uRtrYMQZFgAJ to reproduce.

Broker crashes because of an infinite recursion when creating the frames to deliver the message. This happens only with an empty binary payload (`[<<>>]`), which is easy to create with the management UI. Other clients (e.g. Java client) may create empty list payloads (`[]`), which doesn't trigger the problem.

acogoluegnes:
Fixes #299


michaelklishin:
Backported to `v3.7.x`.

PR #307, opened 2019-02-27
Fix rabbit_pbe for OTP-22

essen:
Some new ciphers have been added. Others will be added once
crypto exposes an interface to get cipher information.

I have opened https://bugs.erlang.org/browse/ERL-864 and will work on a patch to make crypto expose this information. This should take only a few hours.


michaelklishin:
Backported to `v3.7.x`.

PR #308, opened 2019-03-19
Make rabbit_pbe work for OTP-22

essen:
This depends on https://github.com/erlang/otp/pull/2186 so should not be merged yet.

A good cleanup of this module should be planned once we support OTP-22+. We can remove more than half of the module when that happens.


essen:
Ready to be reviewed/merged.

michaelklishin:
Backported to `v3.7.x`.

PR #314, opened 2019-03-26
We don't support aes_ctr before OTP-22

essen:
The cipher was not available via the EVP interface before
Erlang/OTP-22.

Note that the ciphers aes_..._ctr were added in OTP-22 so no need to blacklist them.

This is for master. For v3.7.x please cherry-pick the commit from the previous PR and then this commit.


essen:
Relevant commit in OTP-22 that made `aes_ctr` available through this interface: https://github.com/erlang/otp/commit/c7cab680b80ef9220832151ed2c8c23a5d590b8b

michaelklishin:
Backported to `v3.7.x`.

PR #316, opened 2019-03-29
vm_memory_monitor: handle KB as well as kB

kjnilsson:
There are cases (e.g. inside LXC) where the unit reported in /proc/meminfo uses
KB instead of kB. Handle this case.


kjnilsson:
See this mailing list thread: https://groups.google.com/forum/#!topic/rabbitmq-users/qAh5sn26TOc

Issue #315, opened 2019-03-28
Make decompose_pid compatible with OTP 22 even when compilation happens on OTP 20.3


michaelklishin:
See https://github.com/rabbitmq/rabbitmq-common/pull/313 for background. Prior art in navigating such breaking Erlang standard library API changes is https://github.com/rabbitmq/rabbitmq-common/pull/42.

PR #317, opened 2019-04-02
Support both PID_EXT and NEW_PID_EXT simultaneously

essen:
Better fix for https://github.com/rabbitmq/rabbitmq-common/issues/313 that closes https://github.com/rabbitmq/rabbitmq-common/issues/315


michaelklishin:
@essen just as a note, I've noticed that `encrypt_decrypt` and `encrypt_decrypt_term` tests fail on OTP 22 again (I rebuilt from master a couple of days ago). It has nothing to do with this PR, of course, simply noticed while QA'ing this.

essen:
OK will look into it next week.

PR #321, opened 2019-04-08
rabbit_misc: Generate a `NEW_PID_EXT`-based binary in compose_pid/4

dumbbell:
The `creation` field might not fit into one byte which makes the `PID_EXT` format unsuitable for this case.

The `NEW_PID_EXT` format is supported since Erlang 19.0, so it is safe to always use it, no matter the value of `creation`, because RabbitMQ 3.7+ requires at least Erlang 19.3+.

References #313.


dumbbell:
Thank you!

michaelklishin:
Backported to `v3.7.x`.

PR #322, opened 2019-04-09
Don't support aes_ctr ciphers until OTP-22+

essen:
We disable them when compiling with Erlang <22 because they do not work, and on 22+ for the time being for consistency. We can support them and stream ciphers once we support only 22+ (see added comment). They are also not terribly interesting for what `rabbit_pbe` is used for so it's not a big loss anyway.


michaelklishin:
👍

michaelklishin:
Backported to `v3.7.x`.

PR #323, opened 2019-04-11
rabbit_ssl_options: Blacklist TLS 1.3 until Erlang client implementation is finished

dumbbell:
Erlang 22 will introduce TLS 1.3, but at the time of this commit, only the server side is implemented. If the Erlang client requests TLS 1.3, the server will accept but the client will either hang or crash.

So for now, just blacklist TLS 1.3 to avoid any issues, even on the server side, just to be safe.

This should be backported to `v3.7.x` if accepted.

[#165214130]

### DIFF
--- a/src/rabbit_ssl_options.erl
+++ b/src/rabbit_ssl_options.erl
@@ -18,8 +18,22 @@
 
 -export([fix/1]).
 
-%% POODLE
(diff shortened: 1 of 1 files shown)


dumbbell:
Yes, I agree. Thank you!

dumbbell:
Backported to `v3.7.x`.

PR #328, opened 2019-06-12
New Lager sink for Shovel

michaelklishin:
Drive-by change that is now part of rabbitmq/rabbitmq-shovel#61.

References rabbitmq/rabbitmq-shovel#60.

[#166627333]


michaelklishin:
Backported to `v3.7.x`.

PR #329, opened 2019-06-13
Make default worker pool size configurable

michaelklishin:
Part of rabbitmq/rabbitmq-server#2030.


michaelklishin:
Backported to `v3.7.x` by @dcorbacho 👍

PR #334, opened 2019-08-16
Replace timer:send_after/3 by erlang:send_after/3

dcorbacho:
See http://erlang.org/doc/efficiency_guide/commoncaveats.html#timer-module


michaelklishin:
Backported to `v3.7.x`.

PR #336, opened 2019-10-15
Reserve file handles for quorum queues

dcorbacho:
These should be taken into account into the limits, but always be granted.
Three file handles are reserved on startup for the wal, segment writer and a DETS table,
the rest must be reserved by the queues themselves using `set_reservation/0` or
`set_reservation/1`.


michaelklishin:
Backported to `v3.8.x`.

PR #339, opened 2019-11-12
Move some helpers from rabbitmq-management

michaelklishin:
Part of rabbitmq/rabbitmq-management#749.


dcorbacho:
@michaelklishin The `amqp_table` functions are part of the management agent, are we duplicating them or moving them? There isn't a PR in the agent repo

michaelklishin:
Looking at forwarding those functions. Some had to slightly deviate but some can be forwarded.

michaelklishin:
Backported to `v3.8.x`.

PR #347, opened 2019-12-24
optimisation for 'delegate'

tomyouyou:
%% delegate is an alternative way of doing remote calls. Compared to
%% the rpc module, it reduces inter-node communication. For example,
%% if a message is routed to 1,000 queues on node A and needs to be
%% propagated to nodes B and C, it would be nice to avoid doing 2,000
%% remote casts to queue processes.
Above description is copied from delegate.erl.

The PR is based on the following considerations:
1. It does not affect the operation that must go through the 'delegate', for example：
    {Mem, _BadNodes} = delegate:invoke(Pids, {erlang, process_info, [memory]}) or
    {Results, Errors} = delegate:invoke(MemberPids, ?DELEGATE_PREFIX, FunOrMFA)
2. If there is only one PID for each node of the target, the safe_invoke is directly called without passing through the 'delegate'.

### DIFF
--- a/src/delegate.erl
+++ b/src/delegate.erl
@@ -90,11 +90,23 @@ start_link(Name, Num) ->
     Name1 = delegate_name(Name, Num),
     gen_server2:start_link({local, Name1}, ?MODULE, [Name1], []).
 
+invoke(Pid, FunOrMFA = {gen_server2, _F, _A}) ->  %% optimisation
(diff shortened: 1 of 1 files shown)


michaelklishin:
The delegate mechanism is to avoid extra inter-node traffic for "multi-queue fanouts", that's true. I'll assume that the PR was closed because you've figured it needs more work.

PR #348, opened 2019-12-25
optimisation for 'delegate'

tomyouyou:
%% delegate is an alternative way of doing remote calls. Compared to
%% the rpc module, it reduces inter-node communication. For example,
%% if a message is routed to 1,000 queues on node A and needs to be
%% propagated to nodes B and C, it would be nice to avoid doing 2,000
%% remote casts to queue processes.
Above description is copied from delegate.erl.

The PR is based on the following considerations:

1. It does not affect the operation that must go through the 'delegate', for example：
{Mem, _BadNodes} = delegate:invoke(Pids, {erlang, process_info, [memory]}) or
{Results, Errors} = delegate:invoke(MemberPids, ?DELEGATE_PREFIX, FunOrMFA)
2. If there is only one PID for each node of the target, the safe_invoke is directly called without passing through the 'delegate'.

### DIFF
--- a/src/delegate.erl
+++ b/src/delegate.erl
@@ -90,11 +90,23 @@ start_link(Name, Num) ->
     Name1 = delegate_name(Name, Num),
     gen_server2:start_link({local, Name1}, ?MODULE, [Name1], []).
 
+invoke(Pid, FunOrMFA = {gen_server2, _F, _A}) when is_pid(Pid) ->  %% optimisation
+    case safe_invoke(Pid, FunOrMFA) of
+        {ok,    _, Result} -> Result;
+        {error, _, {Class, Reason, StackTrace}}  -> 
+            erlang:raise(Class, Reason, StackTrace)
+    end;
 invoke(Pid, FunOrMFA) ->
     invoke(Pid, ?DEFAULT_NAME, FunOrMFA).
 
 invoke(Pid, _Name, FunOrMFA) when is_pid(Pid) andalso node(Pid) =:= node() ->
     apply1(FunOrMFA, Pid);
+invoke(Pid, ?DEFAULT_NAME, FunOrMFA = {gen_server2, _F, _A}) when is_pid(Pid) ->  %% optimisation
+    case safe_invoke(Pid, FunOrMFA) of
+        {ok,    _, Result} -> Result;
+        {error, _, {Class, Reason, StackTrace}}  -> 
+            erlang:raise(Class, Reason, StackTrace)
+    end;
 invoke(Pid, Name, FunOrMFA) when is_pid(Pid) ->
     case invoke([Pid], Name, FunOrMFA) of
         {[{Pid, Result}], []} ->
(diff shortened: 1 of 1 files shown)


src/delegate.erl
```diff
@@ -200,6 +235,16 @@ group_pids_by_node(Pids) ->
                  node(Pid), fun (List) -> [Pid | List] end, [Pid], Remote)}
       end, {[], maps:new()}, Pids).
 
+group_local_call_pids_by_node(Pids) ->
+    {LocalPids0, Grouped0} = group_pids_by_node(Pids),
+    maps:fold(fun(K, V, {AccIn, MapsIn}) -> 
+        case V of
+            %% just one Pid for the node
+            [S_v] -> {[S_v | AccIn], MapsIn};
```
michaelklishin: `S_v` uses an unorthodox naming of bindings (variables) which you won't find used elsewhere in RabbitMQ.

src/delegate.erl
```diff
@@ -200,6 +235,16 @@ group_pids_by_node(Pids) ->
...
+            _ -> {AccIn, maps:update_with(K, fun(V1) -> V1 end, V, MapsIn)}
```
michaelklishin: I'm not sure why this update is necessary. If the value is a list of more than one pid, should the clause simply continue?

src/delegate.erl
```diff
@@ -167,6 +193,9 @@ invoke_no_result(Pid, FunOrMFA) when is_pid(Pid) ->
     ok;
 invoke_no_result([], _FunOrMFA) -> %% optimisation
     ok;
+invoke_no_result([Pid], FunOrMFA = {gen_server2, _F, _A}) when is_pid(Pid) -> %% optimisation
```
michaelklishin: Why is `gen_server2` hardcoded here?

src/delegate.erl
```diff
@@ -90,11 +90,23 @@ start_link(Name, Num) ->
     Name1 = delegate_name(Name, Num),
     gen_server2:start_link({local, Name1}, ?MODULE, [Name1], []).
 
+invoke(Pid, FunOrMFA = {gen_server2, _F, _A}) when is_pid(Pid) ->  %% optimisation
+    case safe_invoke(Pid, FunOrMFA) of
+        {ok,    _, Result} -> Result;
+        {error, _, {Class, Reason, StackTrace}}  -> 
+            erlang:raise(Class, Reason, StackTrace)
```
michaelklishin: We do not use `erlang:raise/3` (or `exit/1`, or `throw/1`) in this module. Are you sure that it is safe to just terminate the caller?

michaelklishin:
It would also be very nice to see what kind of measurable improvements this leads to. Have you observed delegate process congestion in metrics of sorts? What kind of difference does this change make there? And for your real world RabbitMQ client workload throughput?

Collecting some numbers would help both convince our team and for you to prove that the change indeed does what it promises to do.

michaelklishin:
(PR closed because this branch contains rejected changes and reverts, not because the idea is not something we would consider; specific feedback was submitted above)

src/delegate.erl
tomyouyou: S_v  is Single_value

src/delegate.erl
tomyouyou: _ -> {AccIn, maps:update_with(K, fun(V1) -> V1 end, V, MapsIn)}  ：  If the value is a list of more than one pid,  the (K,V) will be put to the new map which will be called through delegate.

src/delegate.erl
tomyouyou: "erlang:raise/3" is used because it is copied from the original function "invoke(Pid, Name, FunOrMFA) when is_pid(Pid)"

src/delegate.erl
tomyouyou: It does not affect the operation that must go through the 'delegate', for example：
{Mem, _BadNodes} = delegate:invoke(Pids, {erlang, process_info, [memory]})

In this case, you have to go through an  'delegate'.  "gen_server2 hardcoded here" is to limit modification impact to sending messages

michaelklishin: In other words, we only want to optimize for `gen_server2` processes?

src/delegate.erl
tomyouyou: Could I make the following changes and resubmit the PR?
1. Fork a new branch without any rejected changes and reverts
2. Change  "S_v"  to  "SinglePid"

michaelklishin: Sure. Do you have any contention and/or throughput or latency numbers to share?

michaelklishin: Also, the above comment about the use of `erlang:raise/3` should be addressed. This module's API at the moment does not throw errors.

PR #351, opened 2019-12-31
To alarm when the Erlang process memory is greater than the free memory

tomyouyou:
%% In practice Erlang shouldn't be allowed to grow to more than a half
%% of available memory. The pessimistic scenario is when the Erlang VM
%% has a single process that's consuming all memory. In such a case,
%% during garbage collection, Erlang tries to allocate a huge chunk of
%% continuous memory, which can result in a crash or heavy swapping.
Copied from vm_memory_monitor.erl

So if memory the Erlang process consumed is greater than the free memory of the OS, the Erlang
GC might not work.

The PR only deals with the Linux OS at present.

### DIFF
--- a/src/rabbit_misc.erl
+++ b/src/rabbit_misc.erl
@@ -86,6 +86,7 @@
 -export([report_default_thread_pool_size/0]).
 -export([get_gc_info/1]).
 -export([group_proplists_by/2]).
+-export([try_do/1, try_do/2]).
 
 %% Horrible macro to use in guards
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you but we won't be changing when the alarm mechanism is triggered. It can be perfectly reasonable and expected by the operator for a node to use more RAM than is free, say, when the free memory watermark uses a high value such as `0.8` or even more.

PR #357, opened 2020-02-26
rabbit_env: Add `$RABBITMQ_CONFIG_FILES`

dumbbell:
It is used to specify additional configuration files when using the Cuttlefish format.

Required by rabbitmq/rabbitmq-server#2258.


gerhard:
This is so cool! It will make it super easy to add configs from different sources. Trying it out now.

dumbbell:
Note that the actual feature is implemented in rabbitmq/rabbitmq-server#2258.

This patch just reads the environment variable or sets a default value.

PR #360, opened 2020-03-06
Bump minimum requirement to Erlang/OTP 21.3

dumbbell:
We already announced and document that RabbitMQ 3.7.x requires Erlang/OTP 21.3.

This commit makes it effective. This allows us to use the new `try Class:Reason:Stacktrace` syntax introduced in Erlang/OTP 21.0 and unbreak RabbitMQ 3.7.x with the upcoming Erlang/OTP 23.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ env:
 elixir:
   - '1.9'
 otp_release:
-  - '20.3'
+  - '21.3'
   - '22.2'
 
 install:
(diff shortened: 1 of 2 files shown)


dumbbell:
`v3.7.x`'s rabbitmq-erlang-client fails with Erlang 23 because we still use `erlang:get_stacktrace()` there. We already announced and documented that 3.7.x requires 21.3. Is this ok if I make it true in the implementation by really bumping the requirement to 21.3 (and start to use `try C:R:S`)?

michaelklishin:
Erlang 20.3 has been documented as unsupported since `3.7.19` (October 2019). In addition, Erlang 21.x and 22.x availability for older distributions has been a solved problem for much longer.

So I'd say yes, we should finally actually require 21.3 and add a `3.7.25` release note that mentions that nodes won't even start on 20.3 any more.

dumbbell:
Thank you. I will move forward with this, then.

PR #368, opened 2020-03-20
Add `worker_pool:dispatch_sync` function

pjk25:
`dispatch_sync` sits in between the behavior of `submit` and `submit_async`, blocking the caller until a worker begins the task (`submit` blocks until the task is scheduled and complete, while `submit_async` does not block at all).

This is useful when you want to throttle submissions to the pool from a single process, such that all workers are busy, but there exists no backlog of work for the pool.

### DIFF
--- a/test/worker_pool_SUITE.erl
+++ b/test/worker_pool_SUITE.erl
@@ -80,7 +80,7 @@ dispatch_async_blocks_until_task_begins(Config) ->
                  after 250 ->
                          none
                  end,
-    ?assert(is_pid(SomeWorker), "Dispatched tasks should be running"),
+    ?assert(is_process_alive(SomeWorker), "Dispatched tasks should be running"),
     Pid = spawn(
             fun() ->
                     ok = worker_pool:dispatch_sync(PoolName,
--- a/test/worker_pool_SUITE.erl
+++ b/test/worker_pool_SUITE.erl
@@ -1,69 +1,195 @@
 %% The contents of this file are subject to the Mozilla Public License
 %% Version 1.1 (the "License"); you may not use this file except in
-%% compliance with the License. You may obtain a copy of the License at
-%% https://www.mozilla.org/MPL/
+%% compliance with the License. You may obtain a copy of the License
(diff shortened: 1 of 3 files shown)


test/worker_pool_SUITE.erl
```diff
@@ -0,0 +1,105 @@
...
+             end,
+
+    ok = worker_pool:dispatch_sync(PoolName, Waiter),
+    ok = worker_pool:dispatch_sync(PoolName, Waiter),
+    SomeWorker = receive
+                     {register, WPid} -> WPid
+                 after 250 ->
+                         none
+                 end,
+    ?assert(is_pid(SomeWorker), "Dispatched tasks should be running"),
```
michaelklishin: `is_pid/1` will happily return `true` for a process that's no longer running. Use `is_process_alive/1`.

pjk25: I used `is_pid/1` just to ensure the `receive` above did not return `none` from the timeout, but I think `is_process_alive/1` is an improvement.

pjk25:
> There is already a worker pool suite in the server repo. These tests should be moved there or that suite should be moved to this repo but we should not keep both.

I'll move that suite here, since I think next to it's implementation is a more obvious location (I looked for the suite before creating this one, but didn't find it).

michaelklishin:
The new tests depend on `rabbit_ct_helpers` which was not listed as a test dependency in this repo, hence the CI failures. It's been added in master.

pjk25:
Thanks @michaelklishin

gerhard:
@michaelklishin I need the `worker_pool_dispatch_sync` branch & `startup_memory_fix` from https://github.com/rabbitmq/rabbitmq-server/pull/2279 for [rabbitmq/tgir#8](https://github.com/rabbitmq/tgir/pull/8). Let's keep them around for the duration of TGIR Season 1 which runs through to 25th December 2020. Thanks!

dumbbell:
Backported to 3.8.x as part of rabbitmq/rabbitmq-common#387.

PR #373, opened 2020-04-16
rabbitmq-tools.mk: New `rabbitmq-deps.mk` target

dumbbell:
It generates a file called `$(PROJECT)-rabbitmq-deps.mk` which has a dependency definition line of the form expected by Erlang.mk, for each RabbitMQ component the project depends on.

Therefore the line indicates:
* `git` as the fetch method
* the repository URL
* the Git commit hash the dependency is on

Here is an example for rabbitmq-server:

    dep_rabbit_common := git https://github.com/rabbitmq/rabbitmq-common.git d9ccd8d9cdd58310901f318fed676aff59be5afb
    dep_rabbitmq_cli := git https://github.com/rabbitmq/rabbitmq-cli.git f6eaae292d27da4ded92b7c1b51a8ddcfefa69c2
    dep_rabbitmq_codegen := git https://github.com/rabbitmq/rabbitmq-codegen.git 65da2e86bd65c6b6ecd48478ab092721696bc709

### DIFF
--- a/mk/rabbitmq-tools.mk
+++ b/mk/rabbitmq-tools.mk
@@ -396,3 +396,33 @@ clean-ct-logs-archive::
 	$(gen_verbose) rm -f $(PROJECT)-ct-logs-*.tar.xz
 
 clean:: clean-ct-logs-archive
+
+# --------------------------------------------------------------------
+# Generate a file listing RabbitMQ component dependencies and their
+# Git commit hash.
+# --------------------------------------------------------------------
+
+.PHONY: rabbitmq-deps.mk clean-rabbitmq-deps.mk
+
+rabbitmq-deps.mk: $(PROJECT)-rabbitmq-deps.mk
+	@:
+
+closing_paren := )
+
+define rmq_deps_mk_line
+dep_$(1) := git $(dir $(RABBITMQ_UPSTREAM_FETCH_URL))$(call rmq_cmp_repo_name,$(1)).git $$(git -C "$(2)" rev-parse HEAD)
+endef
+
+$(PROJECT)-rabbitmq-deps.mk: $(ERLANG_MK_RECURSIVE_DEPS_LIST)
+	$(gen_verbose) cat $(ERLANG_MK_RECURSIVE_DEPS_LIST) | \
+	while read -r dir; do \
+	  component=$$(basename "$$dir"); \
+	  case "$$component" in \
+	  $(foreach component,$(RABBITMQ_COMPONENTS),$(component)$(closing_paren) echo "$(call rmq_deps_mk_line,$(component),$$dir)" ;;) \
+	  esac; \
+	done > $@
+
+clean:: clean-rabbitmq-deps.mk
+
+clean-rabbitmq-deps.mk:
+	$(gen_verbose) rm -f $(PROJECT)-rabbitmq-deps.mk


mk/rabbitmq-tools.mk
```diff
@@ -390,3 +390,32 @@ clean-ct-logs-archive::
...
+
+$(PROJECT)-rmq-deps.mk: $(ERLANG_MK_RECURSIVE_DEPS_LIST)
+	$(gen_verbose) :> $@
+	$(verbose) cat $(ERLANG_MK_RECURSIVE_DEPS_LIST) | \
+	while read -r dir; do \
+	  component=$$(basename "$$dir"); \
+	  case "$$component" in \
+	  $(foreach component,$(RABBITMQ_COMPONENTS),$(component)$(closing_paren) echo "$(call rmq_deps_mk_line,$(component),$$dir)" ;;) \
+	  esac; \
+	done > $@
```
gerhard, line 416: Did you mean `>>` here ?

mk/rabbitmq-tools.mk
```diff
@@ -390,3 +390,32 @@ clean-ct-logs-archive::
 	$(gen_verbose) rm -f $(PROJECT)-ct-logs-*.tar.xz
 
 clean:: clean-ct-logs-archive
+
+# --------------------------------------------------------------------
+# Generate a file listing RabbitMQ component dependencies and their
+# Git commit hash.
+# --------------------------------------------------------------------
+
+rmq-deps.mk: $(PROJECT)-rmq-deps.mk
```
gerhard, line 399: Can we call this `deps.mk` instead? If not, then my second proposal is `rabbitmq-server-deps.mk`.

gerhard:
Looking at this target, I am wondering whether we should leverage `query-deps` instead: https://github.com/ninenines/erlang.mk/pull/875 . The current `rmq-deps.mk` implementation looks closer to `query-deps` than this: https://github.com/rabbitmq/rabbitmq-common/blob/a69166ec023bbbcaf027b01c511f276cf557a9eb/mk/rabbitmq-tools.mk#L39-L43

mk/rabbitmq-tools.mk
michaelklishin: `server-deps.mk` should also be descriptive enough?

mk/rabbitmq-tools.mk
dumbbell: `>` is fine, this is the output of the entire `while` loop. I can probably get rid of the initial `:> $@` which is a left-over from previous trials of the implementation.

mk/rabbitmq-tools.mk
dumbbell: The file lists the RabbitMQ component dependencies only, not all dependencies, thus the `rmq` in the name.

The file is named `$(PROJECT)-rmq.deps.mk`, not `rmq-deps.mk`. The latter is just a convenient shortcut to call the recipe. `server` does not make sense in this recipe because it will be executed for rabbitmq-common, the CLI and plugins in the future.

gerhard: In this case `$(PROJECT).deps.mk` feels appropriate.

gerhard: Ok, then `$(PROJECT).rabbitmq.deps.mk` makes most sense.

dumbbell:
If we use `query-deps`, we still have to filter the output (to only keep RabbitMQ components) and format the output (if we want to produce a Makefile).

mk/rabbitmq-tools.mk
dumbbell: Do you mind if we use dashes (`$(PROJECT)-rabbitmq-deps.mk`)? We never used dots so far in the middle of the files we generate.

gerhard:
In which case let's just stick with this and not use `query-deps` for this purpose.

> If we use `query-deps`, we still have to filter the output (to only keep RabbitMQ components) and format the output (if we want to produce a Makefile).

PR #386, opened 2020-05-06
rabbit_env: Add get_used_env_vars/0 to fetch variables used by rabbit_env

dumbbell:
... and their value.

Both prefixed and non-prefixed variables are returned by this function.

While here, fix a conflict between `$RABBITMQ_HOME` and `$HOME` in `var_is_used/1`: the latter shouldn't be considered as used.


michaelklishin:
Thank you!

michaelklishin:
Backported to `v3.8.x`.

PR #390, opened 2020-05-22
Update type specs in rabbit_pbe

michaelklishin:
In follow-up to rabbitmq/rabbitmq-server#2351, #389.


michaelklishin:
Backported to `v3.8.x`.

PR #391, opened 2020-05-30
Introduce rabbit_types:error/2

michaelklishin:
Part of rabbitmq/rabbitmq-server#2356.


michaelklishin:
Backported to `v3.8.x`.

PR #394, opened 2020-06-03
Add a specific shutdown timeout macro and property for message stores

pjk25:
The default timeout of 30 seconds was not sufficient to allow graceful shutdown of a message store with millions of persistent messages. Rather than increase the timeout in general, introduce a new macro with a default of 600 seconds


michaelklishin:
Backported to `v3.8.x`.

PR #395, opened 2020-06-05
Introduce a way to list available NICs in a more presentable format

michaelklishin:
Part of rabbitmq/rabbitmq-cli#424


michaelklishin:
Backported to `v3.8.x`.

PR #397, opened 2020-06-11
Per user connection and channel limits definitions and types

Ayanda-D:
Part of: https://github.com/rabbitmq/rabbitmq-server/pull/2380

**This work has be carried out by** [Erlang Solutions](https://www.erlang-solutions.com/) **and sponsored by** [CloudAMQP](https://www.cloudamqp.com/)

### DIFF
--- a/include/rabbit.hrl
+++ b/include/rabbit.hrl
@@ -25,7 +25,8 @@
     %% password hashing implementation module,
     %% typically rabbit_password_hashing_* but can
     %% come from a plugin
-    hashing_algorithm}).
+    hashing_algorithm,
+    limits = #{}}).
 -record(permission, {configure, write, read}).
 -record(user_vhost, {username, virtual_host}).
 -record(user_permission, {user_vhost, permission}).
--- a/include/rabbit.hrl
+++ b/include/rabbit.hrl
@@ -159,13 +159,21 @@
 
 %% used to track connections across virtual hosts
 %% so that limits can be enforced
--record(tracked_connection_per_vhost,
-    {vhost, connection_count}).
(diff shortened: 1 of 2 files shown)


michaelklishin:
@Ayanda-D this PR cannot be accepted as is as it will break mixed version clusters. Should we expect a follow-up from ESL that makes the user record versioned? We can finish this work ourselves but this can take a while.

Ayanda-D:
@michaelklishin hello, understood and thanks for the feedback. Some delays on our side, ESL have have now commenced work on this PR. We'll be pushing code updates later this week for mixed version cluster compatibility 👍

include/rabbit.hrl
```diff
@@ -243,3 +261,7 @@
 %% Store metadata in the trace files when message tracing is enabled.
 -define(LG_INFO(Info), is_pid(whereis(lg)) andalso (lg ! Info)).
 -define(LG_PROCESS_TYPE(Type), ?LG_INFO(#{process_type => Type})).
+
+%% Execution timeout of connection and channel tracking operations
+-define(TRACKING_EXECUTION_TIMEOUT,
+        rabbit_misc:get_env(rabbit, tracking_execution_timeout, 5000)).
```
michaelklishin, line 267: We have seen too many times before that the 5 second timeout is too prone to false positives on a heavily loaded system.

michaelklishin (approved):

michaelklishin:
This has been merged manually with a merge commit. Keep this in mind when backporting.

michaelklishin:
Backported to `v3.8.x`.

PR #399, opened 2020-06-23
Add a record for node maintenance state

michaelklishin:
Part of rabbitmq/rabbitmq-server#2321.


michaelklishin:
Backported to `v3.8.x`.

PR #400, opened 2020-07-06
rabbit_log: Add Logger handler's log/2 API

dumbbell:
This allows RabbitMQ to configure `rabbit_log` as a Logger handler.

See a related commit in rabbit_prelaunch_early_logging in rabbitmq-server, where `rabbit_log` is being configured as a Logger
handler. The commit message explains the reason behind this.

References rabbitmq/rabbitmq-server#2401.

### DIFF
--- a/src/rabbit_log.erl
+++ b/src/rabbit_log.erl
@@ -16,7 +16,7 @@
 
 -module(rabbit_log).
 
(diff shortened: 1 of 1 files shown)


gerhard:
Back-ported to <kbd>v3.8.x</kkbd>

PR #419, opened 2020-10-20
Update Cowboy to 2.8.0

essen:
This will enable stacktraces in the logs again.

The `rabbitmq-components.mk` file will need to be updated in all repositories. /cc @dumbbell 

I have tested this against auth_backend_http, auth_backend_oauth2, management, web_dispatch, web_mqtt, web_stomp and everything is green. Therefore I do not think there needs to be any code change to upgrade. /cc @michaelklishin


michaelklishin (approved):

michaelklishin:
Will backport later.

michaelklishin:
Backported to `v3.8.x`.

